PR #6, opened 2022-04-22
Delete gitops lib. Refactor intial build controller.

mmorhun:

This PR:
- Deletes GitOps lib and use one from Application Service as it deals with and owns the gitops repository of the component.
- Refactors (rewrites) initial build controller, so now it submits only initial build for components and stops submitting following. It's supposed that all following builds should be triggered by Trigger Template of the component. Any changes in the Component CR must be reflected by Application Service in GitOps repository and then synced by Argo CD.
- Refactors all integration tests of the initial build controller.

Depends on: https://github.com/redhat-appstudio/application-service/pull/96
After merge of the PR above, dependency references to `application-service` must be updated before merging this PR.


controllers/component_build_controller.go
```diff
@@ -100,124 +101,46 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	if err := r.Client.Get(ctx, existingTriggerTemplateNamespacedName, existingTriggerTemplate); err != nil {
-		if errors.IsNotFound(err) {
-			// Build has never been done or cleaned up. Rebuild.
-			log.Info("Previous trigger template not found, rebuild.")
-			return true, nil
+	if err := r.SubmitNewBuild(ctx, component); err != nil {
+		// Try to revert the annotation
+		if err := r.Client.Get(ctx, req.NamespacedName, &component); err == nil {
+			component.Annotations[InitialBuildAnnotationName] = "false"
+			if err := r.Client.Update(ctx, &component); err != nil {
```
brunoapimentel, line 122: What do you think about setting the `InitialBuildAnnotation` to `true` here instead of in L114? I'm thinking this could avoid us keeping the revert logic.

controllers/component_build_controller.go
```diff
@@ -100,124 +101,46 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-// The criterion is equality of existing and expected trigger template of the component.
-func (r *ComponentBuildReconciler) IsNewBuildRequired(ctx context.Context, component appstudiov1alpha1.Component) (bool, error) {
-	log := r.Log.WithValues("Namespace", component.Namespace, "Application", component.Spec.Application, "Component", component.Name)
-
-	expectedTriggerTemplate, err := gitops.GenerateTriggerTemplate(component)
-	if err != nil {
-		return false, err
+	// Set initial build annotation to prevent next builds
+	component.Annotations[InitialBuildAnnotationName] = "true"
+	if err := r.Client.Update(ctx, &component); err != nil {
```
brunoapimentel, line 114: On a more philosophical note, I'm wondering if we should really update the Component CR here (even if it's only an annotation). I understood that the idea was to keep all the Component updates only in HAS. 

Are we opening up an exception for annotations?

controllers/component_build_controller.go
mmorhun: This is done on purpose. If we fail to annotate component CR, another build will be submitted. That shouldn't happen.

PR #11, opened 2022-05-03
Allow the overriding of the default Tekton bundle

brunoapimentel:
PLNSRVCE-141

Allows the Tekton bundle used in the created PipelineRuns to be overriden.
Relates to https://github.com/redhat-appstudio/application-service/pull/106.


Michkov (approved):

controllers/component_build_controller_test.go
```diff
@@ -246,6 +263,12 @@ var _ = Describe("Component initial build controller", func() {
 			}
 			Expect(k8sClient.Create(ctx, gitSecret)).Should(Succeed())
 
+			// Configure the build bundle
```
mmorhun, line 266: I'd create a separate test Context where only build bundle cases (3 of them) are tested in a separate test cases.

PR #14, opened 2022-05-06
Workaround: create PVC for initial build to unblock pipeline run

mmorhun:

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -23,6 +23,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


controllers/component_build_controller.go
```diff
@@ -142,6 +143,26 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component appstudiov1alpha1.Component) error {
 	log := r.Log.WithValues("Namespace", component.Namespace, "Application", component.Spec.Application, "Component", component.Name)
 
+	// TODO delete this block which is workaround for delayed sync of pvc
+	workspaceStorage := gitops.GenerateCommonStorage(component, "appstudio")
+	existingPvc := &corev1.PersistentVolumeClaim{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: workspaceStorage.Name, Namespace: workspaceStorage.Namespace}, existingPvc); err != nil {
```
gabemontero, line 149: Since we have a potential race condition with the delayed sync, we probably want to employ the use of https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate around the Get and Create here

controllers/component_build_controller.go
```diff
@@ -142,6 +143,26 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		if errors.IsNotFound(err) {
+			// Patch PVC size to 1 Gi, because default 10 Mi is not enough
+			workspaceStorage.Spec.Resources.Requests["storage"] = resource.MustParse("1Gi")
+			// Create PVC (Argo CD will patch it later)
+			err = r.Client.Create(ctx, workspaceStorage)
+			if err != nil {
```
gabemontero, line 155: Then within the PollImmediate, if the error here is "already exists" we just move on

Otherwise we retry, in case there are some intermittent error.

Then if things keep on failing for a few seconds, we give up and return the unable to create common storage

mmorhun:
@gabemontero thank you for sharing your experience. ACK.
For this particular case, we have delay about 5 mins, so no race condition should happen.
When the bug fixed and the PVC is in place, the operator just skips the block as PVC exists (in the worst case, an error will be logged and a new reconcile will skip the block).

PR #17, opened 2022-05-12
Do not use cache when interactinig with git repository secret

mmorhun:

On a cluster with huge number of secrets default cache system of operator causes OOM.
As the secret with git credentials has no obligations and initial build happens once per component, we can use non-caching k8s client to avoid this issue.

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -46,9 +46,10 @@ const (
 
 // ComponentBuildReconciler watches AppStudio Component object in order to submit builds
 type ComponentBuildReconciler struct {
-	client.Client
-	Scheme *runtime.Scheme
-	Log    logr.Logger
+	Client           client.Client
+	NonCachingClient client.Client
+	Scheme           *runtime.Scheme
+	Log              logr.Logger
 }
 
 // SetupWithManager sets up the controller with the Manager.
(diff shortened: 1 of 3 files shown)


AndrienkoAleksandr (approved):

gabemontero:
@mmorhun @Michkov the pattern of using both a watch / shared informer / caching client (which you all are obviously aware of from what this PR is dealing with) and a direct to the API server client is something I both employed myself and seen other employ with controllers, though typically with the k8s shared informer API based controllers, vs. the controller runtime based API (which uses the shared informer stuff under the covers). 

So certainly if this problem is causing a lot of disruption, I would +1 the approach here for now.

That said, @mmorhun are you aware of / did you look into https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md ?

The more elegant / longer term solution might be to filter the secret cache.  If we have time to pursue that with this PR, that would be good.  But a follow up PR / Jira seems reasonable too.

mmorhun:
@gabemontero thank you for the comment :+1: . I used cache filtering before, however, it's not possible to use it in this situation, because the secret doesn't have any requirements (to have a valid selector) and could be in any namespace.

gabemontero:
> @gabemontero thank you for the comment +1 . I used cache filtering before, however, it's not possible to use it in this situation, because the secret doesn't have any requirements (to have a valid selector) and could be in any namespace.

cool deal  - thanks for the confirmation / detail @mmorhun

PR #20, opened 2022-05-13
test

psturc:


psturc:
Sorry about the noise guys! We were testing if we could get rid of openshift-ci bot's comments (unfortunately it seems that we can't, since it's enabled on the [redhat-appstudio](https://github.com/orgs/redhat-appstudio/people?query=openshift) org level). I'm closing this PR.

PR #23, opened 2022-05-15
Update test version of application-service

stuartwdouglas:


stuartwdouglas:
Are these tests actually run on CI? This would work for local development if you still have the old version in your GOPATH, but it should have failed on CI.

mmorhun (approved):

PR #24, opened 2022-05-16
Add new component image controller

mmorhun:
#### What does this MR do?

This MR introduces a new controller that watches all Component related PipelineRuns and if any finish successfully, then checks for a newly built component image and updates corresponding component configuration.

#### How to test this MR?

1. Create an Application
2. Create a Component of the Application
3. Wait until initial build succeeds
4. Trigger a new build via the Component webhook
5. Wait until the webhook build finishes
6. Check the Component, it should contain the new image under `Spec.Build.ContainerImage`

### DIFF
--- a/controllers/component_image_controller.go
+++ b/controllers/component_image_controller.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	"knative.dev/pkg/apis"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -46,15 +47,16 @@ const (
 	BuildImageTaskName      = "build-container"
 )
 
-// NewComponentImageReconciler reconciles a Frigate object
-type NewComponentImageReconciler struct {
+// ComponentImageReconciler reconciles a Frigate object
+type ComponentImageReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
-	Log    logr.Logger
+	Scheme        *runtime.Scheme
+	Log           logr.Logger
+	EventRecorder record.EventRecorder
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *NewComponentImageReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *ComponentImageReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&tektonapi.PipelineRun{}, builder.WithPredicates(predicate.Funcs{
 			CreateFunc: func(e event.CreateEvent) bool {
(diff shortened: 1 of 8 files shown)


controllers/component_image_controller.go
```diff
@@ -0,0 +1,248 @@
...
+					return false
+				}
+
+				if old.Status.CompletionTime != nil {
+					// Build had been finished before this update
+					return false
+				}
+
+				// Ensure the PipelineRun belongs to a Component
+				if new.ObjectMeta.Annotations[ComponentAnnotationName] == "" && new.ObjectMeta.Labels[ComponentAnnotationName] == "" {
```
gabemontero, line 88: for my own education @mmorhun  I looked at https://github.com/redhat-appstudio/application-service/blob/main/gitops/generate_build.go and I see that yeah the annotation gets set at https://github.com/redhat-appstudio/application-service/blob/86571e38f52e6bbe67b945dd58b695af26ae834d/gitops/generate_build.go#L87 as part of generating the trigger template

and the label gets set at https://github.com/redhat-appstudio/application-service/blob/86571e38f52e6bbe67b945dd58b695af26ae834d/gitops/generate_build.go#L87 for the "initial pipeline run"

given those separate paths, I'm "guessing" there is a chance where a problem could occur and one of the two might not be set, and we don't want to process in that case.

Is that the rationale? And that we can't assume if the label is there, the annotation will be, where if we could assume that, we would leverage a label selector during the construction of the cache when building the manager?

controllers/component_image_controller.go
```diff
@@ -0,0 +1,248 @@
...
+	switch len(tasks.Items) {
+	case 0:
+		// Build task not found, it was skipped
+		log.Info(fmt.Sprintf("No new image built in '%v' PipelineRun for '%v' Comoponent", req.NamespacedName, componentKey))
+		return ctrl.Result{}, nil
+	case 1:
+		taskrun = &tasks.Items[0]
+	default:
+		// Should not happen
+		log.Info(fmt.Sprintf("Found %d build tasks for %v PipelineRun", len(tasks.Items), pipelineRun))
```
gabemontero, line 208: feels like we should minimally generate an k8s event for this

*maybe* a k8s metric / alert longer term

gabemontero (commented):
a question and a minor suggestion (which I would block on if you have a strong reason/opinion on it) after an initial pass @mmorhun

yangcao77:
HAS just removed the `Spec.Build.ContainerImage` from component CR. use `spec.containerImage` instead. So if `spec.containerImage != "" && spec.source.git != nil`, it is a build image

controllers/suite_util_test.go
```diff
@@ -0,0 +1,317 @@
...
+	pipelineRuns := &tektonapi.PipelineRunList{}
+	labelSelectors := client.ListOptions{Raw: &metav1.ListOptions{
+		LabelSelector: "build.appstudio.openshift.io/component=" + componentKey.Name,
+	}}
+	err := k8sClient.List(ctx, pipelineRuns, &labelSelectors)
+	Expect(err).ToNot(HaveOccurred())
+	return pipelineRuns
+}
+
+func deleteComponentInitialPipelienRuns(componentKey types.NamespacedName) {
```
brunoapimentel, line 139: Typo in "pipelien".

controllers/suite_util_test.go
```diff
@@ -0,0 +1,317 @@
...
+		apis.Condition{
+			Reason: "Failed",
+			Status: "False",
+			Type:   apis.ConditionSucceeded,
+		},
+	}
+	Expect(k8sClient.Status().Update(ctx, pipelineRun)).Should(Succeed())
+}
+
+func succeedInitialPipelienRun(componentKey types.NamespacedName) {
```
brunoapimentel, line 230: Typo in "pipelien".

controllers/suite_util_test.go
```diff
@@ -0,0 +1,317 @@
+/*
```
brunoapimentel, line 1: Nice, the helper functions look a lot better in their own file. :+1:

controllers/component_image_controller.go
mmorhun: Thanks, implemented

gabemontero (approved):
LGTM

Michkov (approved):

PR #26, opened 2022-05-17
Add integration tests to build bundle configuration

brunoapimentel:
Should be merged after #24

### DIFF
--- a/controllers/component_build_controller_test.go
+++ b/controllers/component_build_controller_test.go
@@ -95,7 +95,6 @@ var _ = Describe("Component initial build controller", func() {
 
 	Context("Check if build objects are created", func() {
 
-		var buildBundle string
 		var gitSecret *corev1.Secret
 
 		BeforeEach(func() {
@@ -107,11 +106,6 @@ var _ = Describe("Component initial build controller", func() {
 				},
 			}
 			Expect(k8sClient.Create(ctx, gitSecret)).Should(Succeed())
-			// Configure the build bundle
-			buildBundle = "quay.io/some-repo/some-bundle:0.0.1"
-			createConfigMap(prepare.BuildBundleConfigMapName, HASAppNamespace, map[string]string{
-				prepare.BuildBundleConfigMapKey: buildBundle,
-			})
 			// Create component that refers to the git secret
 			component := &appstudiov1alpha1.Component{
 				TypeMeta: metav1.TypeMeta{
@@ -142,7 +136,6 @@ var _ = Describe("Component initial build controller", func() {
 			// Clean up
 			deleteComponentInitialPipelineRuns(resourceKey)
 			deleteComponent(resourceKey)
-			deleteConfigMap(prepare.BuildBundleConfigMapName, HASAppNamespace)
 			Expect(k8sClient.Delete(ctx, gitSecret)).Should(Succeed())
 		})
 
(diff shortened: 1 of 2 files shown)


controllers/component_build_controller_test.go
```diff
@@ -193,4 +187,65 @@ var _ = Describe("Component initial build controller", func() {
...
+			deleteConfigMap(configMapKey)
+		})
+
+		It("should use the build bundle specified if a configmap is set in the default bundle namespace", func() {
+			buildBundle := "quay.io/some-repo/some-bundle:0.0.2"
+
+			componentKey := types.NamespacedName{Name: HASCompName, Namespace: HASAppNamespace}
+			configMapKey := types.NamespacedName{Name: prepare.BuildBundleConfigMapName, Namespace: prepare.BuildBundleDefaultNamespace}
+
+			createNamespace(prepare.BuildBundleDefaultNamespace)
```
brunoapimentel, line 220: The test works as it is, but I'd like to delete this namespace to avoid leaving unused stuff behind. When I try to delete it, though, the test times out, probably because the deletion is not working. 

Not sure what I'm doing wrong:
```
func deleteNamespace(name string) {
	resourceKey := types.NamespacedName{Name: name}
	// Delete
	Eventually(func() error {
		f := &corev1.Namespace{}
		k8sClient.Get(ctx, resourceKey, f)
		return k8sClient.Delete(ctx, f)
	}, timeout, interval).Should(Succeed())

	// Wait for delete to finish
	Eventually(func() error {
		f := &corev1.Namespace{}
		return k8sClient.Get(ctx, resourceKey, f)
	}, timeout, interval).ShouldNot(Succeed())
}
```

gabemontero: one thing that can delay namespace deletion is that pod termination within the namespace can get bogged down for any number of reasons; I've seen it happen if Pods mount certain type of volumes for example.

I see a couple of options @brunoapimentel : 
1)  override the delete grace period per the explanation of that field in the `DeleteOptions` i.e.  https://github.com/kubernetes/apimachinery/blob/430b920312ca0fa10eca95967764ff08f34083a3/pkg/apis/meta/v1/types.go#L470-L475 and then use `Delete(ctx context.Context, obj Object, opts ...DeleteOption) error` to get it to process immediately, and let k8s garbage collection proceed in the background
2) if you are really curious about what is delaying things, don't return on the delete, but call it in the background, and the create a poller that inspects Events on the namesapce before you exit;  admittedly a much more involved endeavor :-) .... but if you are curious and need help crafting something like that, let me know

my guess is try 1) first

mmorhun (approved):
LGTM

controllers/component_build_controller_test.go
brunoapimentel: So, setting the grace period to 0 did not work (I remember I had tried it before...)

brunoapimentel: I also tried the force delete instructions from [here](https://stackoverflow.com/questions/55853312/how-to-force-delete-a-kubernetes-namespace#55857926), but no luck.

gabemontero: hmmm ... then my 2. is the only think I can think of to suggest to you @brunoapimentel

brunoapimentel: At this point, I guess it's better to merge this PR as it is, and, since the issue is not affecting the tests, we can fix it in a following PR.

PR #33, opened 2022-05-30
Create pr.yml

psturc:


psturc:
sorry for the noise. I'm trying to figure out why GH actions won't get triggered in PR nr. 30

PR #40, opened 2022-06-24
Initial pipelines-as-code integration

Michkov:
Adds initial implementation of PaC into Build Service.
Currently both Trigger and PaC builds are supported. If the Component has `pipelinesascode: "1"` annotation, then PaC build is used. Otherwise, previous behavior applies.


controllers/component_build_controller.go
```diff
@@ -126,34 +145,176 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+	} else {
+		// Set initial build annotation to prevent next builds
+		component.Annotations[InitialBuildAnnotationName] = "true"
+		if err := r.Client.Update(ctx, &component); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if err := r.SubmitNewBuild(ctx, component); err != nil {
+			// Try to revert the annotation
```
gabemontero, line 170: so does this mean that if we can't build via pipeline as code, we try to build "normally" ?

If so, I think some logging and generating some events in the user's namesapce / workspace would be good here.

controllers/component_build_controller.go
```diff
@@ -126,34 +145,176 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			if err := r.Client.Get(ctx, req.NamespacedName, &component); err == nil {
+				component.Annotations[InitialBuildAnnotationName] = "false"
```
gabemontero, line 172: also, it seems like GeneratePipelineRun creates several annotations.  Should we not clean up all of them if this is in fact a"retry as normal / not as pipeline as code" path.

gabemontero (commented):
a couple of comments around one guess as to what is going on / question on my part @mmorhun

controllers/component_build_controller.go
mmorhun: No, we don't. As we are in switching to PaC phase, we currently support both builds: trigger (old) and PaC (new). The code you asked about is completely related to the trigger build. We use an annotation to mark a component that initial build happened (and some other preparations done). But if something in the preparation or build submition failed, we just delete (revert) the annotation, so next reconcile loop will retry. The annotation is added before all the preparations and build submition because there is a risk that the component is changed and the annotation update will fail, so the initial build would be run twice. To avoid that, we set annotation first and in case of a preparation error (a rare case) just revert it.

brunoapimentel (approved):
The only thing I'd like to point is that we should add tests to cover at least the basics of the newly added github package, as well as add new integration tests to the component_build_controller to cover the new behavior.

This could be done in a next PR, though, this one LGTM.

mmorhun:
Tested manually the flow:
1. Create a component repository (forked from https://github.com/devfile-samples/devfile-sample-go-basic).
2. Installed PaC GitHub Application (see [docs](https://pipelinesascode.com/docs/install/github_apps/)).
3. Created AppStudio Application and Component (with `pipelinesascode: "1"` annotation).
4. Checked for a new PR in the Component repository.
5. Fixed `Dockerfile` path in the PR: from `Dockerfile` to `docker/Dockerfile`.
6. Added `/retest` comment in the PR, it triggered a new on-pr pipeline.
7. When pipeline finished successfully, merged the PR.
8. A new on-push pipeline is triggered.
9. Component deployment is available.

PR #41, opened 2022-07-15
Create ApplicationSnapshot for non-HACBS builds

mmorhun:
This PR creates `ApplicationSnapshot` object when a new image for a `Component` is built.
Note, it creates an `ApplicationSnapshot` on rebuilds (image built, but no need in updating the `Component` as the correct image reference is already set).

Example of created `ApplicationSnapshot`:
```yaml
kind: ApplicationSnapshot
metadata:
  creationTimestamp: "2022-07-20T05:37:19Z"
  generation: 1
  labels:
    build.appstudio.openshift.io/application: application-sample
  name: snapshot-of-application-application-sample-1658295439353896957
  namespace: test
  ownerReferences:
  - apiVersion: appstudio.redhat.com/v1alpha1
    kind: Application
    name: application-sample
    uid: b64eeeb3-ea00-4bca-ac16-6d891a8a2af4
  resourceVersion: "1068921"
  uid: 2fac3dd5-a013-4d83-b804-88f0a39c1bee
spec:
  application: application-sample
  artifacts: {}
  components:
  - containerImage: registry.io/username/foobar:go-srv
    name: application-sample-go-component
  - containerImage: registry.io/username/foobar:dev-code-with-quarkus-29b0823364ba05bd5a9d3a89d4e6cad57d2d3723
    name: component-sample
  displayDescription: 'Snapshot of "application-sample" application created at 2022
    Jul 20 Wednesday 05:37:19. Component images: application-sample-go-component->docker.io/mm4eche/foobar:go-srv
    component-sample->docker.io/mm4eche/foobar:dev-code-with-quarkus-29b0823364ba05bd5a9d3a89d4e6cad57d2d3723'
  displayName: Snapshot of "application-sample" application
```

### DIFF
--- a/controllers/component_image_controller.go
+++ b/controllers/component_image_controller.go
@@ -19,9 +19,12 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/selection"
@@ -36,6 +39,7 @@ import (
 
 	"github.com/go-logr/logr"
 	appstudiov1alpha1 "github.com/redhat-appstudio/application-service/api/v1alpha1"
+	appstudiosharedv1alpha1 "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
 	tektonapi "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 )
 
@@ -46,6 +50,8 @@ const (
 	UpdateComponentAnnotationName = "appstudio.redhat.com/updateComponentOnSuccess"
 	BuildImageTaskName            = "build-container"
 	PullRequestAnnotationName     = "pipelinesascode.tekton.dev/pull-request"
+
+	k8sNameMaxLen = 253
 )
 
 // ComponentImageReconciler reconciles a Frigate object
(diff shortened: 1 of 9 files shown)


controllers/component_image_controller.go
```diff
@@ -245,5 +252,63 @@ func (r *ComponentImageReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 	}
 
+	// Create Application Snapshot
+	// This object will be used to regenerate the GitOps repository
+	// (based on the Binding, which references the Snapshot)
+	applicationSnapshot := generateApplicationSnapshot(&component, imageReference)
+	if err := r.Client.Create(ctx, applicationSnapshot); err != nil {
+		log.Error(err, fmt.Sprintf("Failed to create application snapshot for the new image: %s", imageReference))
+		return ctrl.Result{}, err
+	}
```
maysunfaisal, line 262: So I am not sure if creating a snapshot for a single Component Image is going to work. Here is a justification:

This is how an instance of a binding looks:
```
kind: ApplicationSnapshotEnvironmentBinding
metadata:
  name: applicationsnapshotenvironmentbinding-sample
  labels: 
    appstudio.application: application-sample # Should always match application value below
    appstudio.environment: staging # Should always match env value below
spec:
  application: application-sample
  environment: staging
  snapshot: applicationsnapshot-sample
  components:
  - name: component-sample
    configuration:
      replicas: 3
      env:
      - name: FOO
        value: BAR
  - name: component-sample2
    configuration:
      replicas: 2
      env:
      - name: FOO2
        value: BAR2
```

A binding has all these components listed that belongs to the application. So when it references a snapshot, it is expecting to find the built image for the all listed components in the binding.spec.. So in this instance, component-sample and component-sample2 image needs to be be present in the snapshot like this:

```
kind: ApplicationSnapshot
metadata:
  name: applicationsnapshot-sample
spec:
  application: application-sample
  displayName: my-snapshot
  displayDescription: my first snapshot
  components:
  - name: component-sample
    containerImage: quay.io/redhat-appstudio/user-workload:application-service-system-component-sample
  - name: component-sample2
    containerImage: quay.io/redhat-appstudio/user-workload:application-service-system-component-sample2
```

The way I was hoping this would be solved was to query the other components that belong to an application and copy their built image reference when creating a snapshot. So in our example, if component-sample2 is being built, we have the newly built image but we can find out what the built image reference was for component-sample and create a snapshot for both of them 

What do you think? 🤔

mmorhun: I see, will implement. Thank you for the details.

gabemontero:
/hold cancel

new commits removed my inadvertent application of the lgtm label

controllers/component_image_controller.go
mmorhun: @maysunfaisal updated the PR and description according to the above. Please review.

maysunfaisal (approved):

PR #48, opened 2022-08-08
Pipelines as Code using webhook implementation

mmorhun:

Implements support of Pipelines as Code webhook for GitHub repositories.
Adds stub for other git providers to be supported.

Depends on: https://github.com/redhat-appstudio/application-service/pull/160


controllers/component_build_controller.go
```diff
@@ -152,17 +160,66 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			// Do not reconcile, because configuration must be fixed before it is possible to proceed.
+			return ctrl.Result{}, nil
+		}
+
+		// Expect that the secret contains token for Pipelines as Code webhook configuration,
+		// but under <git-provider>.token field. For example: github.token
+		// Also it can contain github-private-key and github-application-id
+		// in case GitHub Application is used instead of webhook.
+		pacSecret := corev1.Secret{}
+		if err := r.NonCachingClient.Get(ctx, types.NamespacedName{Namespace: pipelinesAsCodeNamespace, Name: gitopsprepare.PipelinesAsCodeSecretName}, &pacSecret); err != nil {
```
gabemontero, line 183: So this is a tricky item, using the non caching client
1) on the side arguing for it, to use the caching client here, you need to set up a watch for secrets in the all the namespaces build could occur in; such watches will create **A LOT** of events
2) on the side arguing against it, using the non caching client means we are hitting the api server on all builds that use pipeline as code; if that becomes the norm, that is **A LOT** of extra activity on the api server (fwiw, in openshift the various controllers in OCP by default are instructed to not use the non caching client because of this)

I'm conflicted, but I think with the absence of testing at scale, this is probably the right initial choice to make, because of the simplicity on our end.  Plus, I honestly do not have a sense of how much the non caching client is being used across app studio.  Perhaps this is a drop in the bucket, and we'll mitigate this back making sure our hosted service has given sufficient processing power to the API server to handle.

But I mention all this, even though I am not asking for a change, just so we document that we have at least considered the pro and cons here @mmorhun

controllers/component_build_controller.go
```diff
@@ -152,17 +160,66 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			r.Log.Error(err, "failed to get Pipelines as Code secret")
```
gabemontero, line 184: an event in `pipelineAsCodeNamespace` seems warranted here as well.

controllers/component_build_controller.go
```diff
@@ -152,17 +160,66 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 			return ctrl.Result{}, err
 		}
-		if err := r.GeneratePullRequest(ctx, component); err != nil {
+
+		if err := validatePaCConfiguration(gitProvider, pacSecret.Data); err != nil {
+			r.Log.Error(err, "Invalid configuration in Pipelines as Code secret")
```
gabemontero, line 189: an event in `pipelineAsCodeNamespace` seems warranted here as well.

controllers/component_build_controller.go
```diff
@@ -152,17 +160,66 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		// Copy or update global PaC configuration in local namespace
+		if err := r.propagatePaCConfigurationSecretToLocalNamespace(ctx, req.Namespace, pacSecret); err != nil {
 			return ctrl.Result{}, err
 		}
 
+		// Generate webhook secret for the component git repository if not yet generated
+		// and stores it in the corresponding k8s secret.
+		webhookSecretString, err := r.ensureWebhookSecret(ctx, component)
+		if err != nil {
+			return ctrl.Result{}, err
```
gabemontero, line 203: I'm assuming that errors here are expected to be intermittent, and simply returning and error and retrying makes sense here

controllers/component_build_controller.go
```diff
@@ -152,17 +160,66 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+
+		if err := r.EnsurePaCRepository(ctx, component, pacSecret.Data); err != nil {
+			return ctrl.Result{}, err
```
gabemontero, line 207: I'm assuming that errors here are expected to be intermittent, and simply returning and error and retrying makes sense here

controllers/component_build_controller.go
```diff
@@ -152,17 +160,66 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+
+		bundle := gitopsprepare.PrepareGitopsConfig(ctx, r.NonCachingClient, component).BuildBundle
+		mrUrl, err := ConfigureRepositoryForPaC(component, pacSecret.Data, webhookTargetUrl, webhookSecretString, bundle)
+		if err != nil {
+			r.Log.Error(err, "failed to setup repository for Pipelines as Code")
+			return ctrl.Result{}, err
```
gabemontero, line 214: I'm assuming that errors here are expected to be intermittent, and simply returning and error and retrying makes sense here

or given you chose to error log, could it be permament, and if so, I would say log an event in the user's namespace as well so they can understand more easily (seeing events in his namespace vs. looking at logs in a namespace the user cannot access

controllers/component_build_controller.go
```diff
@@ -198,71 +255,288 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
-func (r *ComponentBuildReconciler) EnsurePACRepository(ctx context.Context, component appstudiov1alpha1.Component) error {
-	repository := gitops.GeneratePACRepository(component)
-	existingRepository := &pacv1alpha1.Repository{}
-	if err := r.Client.Get(ctx, types.NamespacedName{Name: repository.Name, Namespace: repository.Namespace}, existingRepository); err != nil {
+func (r *ComponentBuildReconciler) getPaCRoutePublicUrl(ctx context.Context) (string, error) {
+	pacWebhookRoute := &routev1.Route{}
+	pacWebhookRouteKey := types.NamespacedName{Namespace: pipelinesAsCodeNamespace, Name: pipelinesAsCodeRouteName}
+	if err := r.Client.Get(ctx, pacWebhookRouteKey, pacWebhookRoute); err != nil {
+		r.Log.Error(err, "failed to get Pipelines as Code route")
```
gabemontero, line 262: probably intermittent or should we issue event ?

gabemontero (changes requested):
some assist users with debugging by submitting events requests, and an aside comment (though not asking for change) around the use of caching client @mmorhun

controllers/component_build_controller.go
mmorhun: @gabemontero thank you for the detailed review.
Let me explain my decision to use non caching client here. The line above is needed to read configured credential for Pipelines as Code. We need them in order to know how (for app or webhook) we should configure component git repository and to do the configuration itself (we need credentials from the secret). As you probably noticed, the PaC secret is not in the component namespace, but in global Pipelines as Code one, so it is related to PaC configuration rather than a component one. Now about reading secrets. If we use caching client, it caches ALL secrets in the cluster which is a huge amount of data. We used to have OOM problems on staging because the operator just tried to read a config map (usually a cluster has a lot of config maps and secrets, in short we'd be in trouble because amount of the operator RAM consumption would depend on cluster size. For our staging it was several gigabytes where normal consumption is one-two hundred megabytes). To overcome such huge caches, we may (and do use for other secrets withing the component namespace) custom cache configuration. But the only way to customize cache is to use a label selector, which means all secrets that are read by the operator must have the specified label. We use such label in local namespace for component related secrets, but we just cannot require the label on the global PaC secret outside the component namespace (that would brought additional burden for users and error prone, but if we could require the label, it would allow us to use only caching client in the implementation above. But in my understanding, it is not fair to ask for a build-service related label in global PaC config namespace).
However, there is a good piece of news: the code above (that uses non caching client) is called only on the component onboarding and not used on PaC builds.
I hopes this clears things behind my decision. Anyway, I am open for any suggestions to improve the code.

In the future, we might have a common label for whole AppStudio and require the label on global PaC secret (as PaC is a part of AppStudio deployment). But this should be done on higher level (I'll rise this topic on one of the next architecture calls).

gabemontero: OK cool @mmorhun you appear on the same page wrt the cons of using the caching client, and detailing your thought process to using the non caching client confirms what I thought that you did not pick that without the sufficient consideration.

And yes, the additional clarification that the piece of code in question is only for component onboarding and not a per build thing is very good news, and IMO confirms this approach is OK  / we are good here.

Leave this open so other see and understand, or mark resolved to shorten your PR discussion length, at your discretion @mmorhun  - thanks

controllers/component_build_controller.go
mmorhun: ok

controllers/component_build_controller.go
mmorhun: ok

controllers/component_build_controller.go
mmorhun: Yes, an error could occur because of network issue or because the resource was changed. Both are handled by a retry.

controllers/component_build_controller.go
mmorhun: Yes, expected errors are intermittent. Permanent error could happen if PaC repository yaml generation fails, but this would be caught by tests.

controllers/component_build_controller.go
mmorhun: I don't think so. We can fail to get the route if PaC is not installed or not ready yet. In that case just retrying should help in waiting for PaC readiness.

controllers/component_build_controller.go
mmorhun: There are a number of reason why it could fail: from network issues to invalid credentials. Ideally, we should have a flag among return values if operator should retry, but I'd leave it for the future improvement.

mmorhun: It worth creating an event to let users know the reason of failure (even if we log many identical events), thank you @gabemontero for pointing this out.

controllers/component_build_controller.go
brunoapimentel: Thanks for bringing this discussion, this was a very informative read.

pkg/github/github-client.go
```diff
@@ -0,0 +1,195 @@
+/*
```
brunoapimentel, line 1: A small nitpick: all files in the project are using snake_case, but github-client and github-helper are using dashes. Not sure if there's a pattern I'm missing here.

mmorhun: You are right, thank you for pointing out. I've renamed the files.

Michkov (approved):
Tested with webhook  integration and also with github app integration, working well

psturc (commented):
just a very minor issue. otherwise I think it'd be worth to check suggestions from the linter (PR check)

pkg/github/github_helper_debug_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+package github
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+)
+
+// THIS FILE IS NOT UNIT TESTS
```
brunoapimentel, line 25: I don't think it's the scope of this PR, but I was wondering if it's worth to create a test user/repository in Github so we could make these tests here work like integration tests. I think they cover the classes in this pkg pretty well.

main.go
```diff
@@ -81,6 +85,11 @@ func main() {
 
 	makeSureTektonCRDsAreInstalled()
 
+	if err := routev1.AddToScheme(scheme); err != nil {
+		setupLog.Error(err, "unable to add triggers api to the scheme")
```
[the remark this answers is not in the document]
mmorhun: Yeah, copy paste, thanks for pointing out.

pkg/github/github_helper_debug_test.go
mmorhun: It's hard to say. Obviously, we cannot use a common repository, only a common template. Then, to run tests, we have to configure/reuse GitHub credentials, clone template repository, clean it up after success or fail. This is not straightforward flow for in-repo tests. I'd leave it for more complex e2e test on CI. Moreover, it's unlikely that interactions with existing GitHub API would change. Even if it happens, it would be a rare case. Similar situation with changes in code: if someone touches parts that related to GitHub interactions it's easy to test and make sure they work as expected. However, I still don't have strong opinion on this. If you think we need such tests - please create an issue.

gabemontero (approved):
/lgtm

pkg/github/github_helper_debug_test.go
brunoapimentel: Yeah, it's definitely a more-complicated-than-usual type of test. I was mostly trying to figure out how to have this part covered with tests, but I do agree that the e2e tests should be a better place to have this functionality covered. So let's leave this as is.

brunoapimentel (approved):

PR #51, opened 2022-08-15
Rework GitHub client to manage PaC PR depending on current state

mmorhun:

This PR adds logic that analyzes current state of the Pipelines as Code configuration in the component repository and updates only needed, if any, parts.
Conditions handled (arbitrary combinations of the below, independent on if the condition true or false):
- configuration is up to date in main branch (no PR is needed)
- branch to create PR from exists
- content of branch to create PR from is up to date
- PR exists

This PR fixes problem with empty or duplicated commits.

### DIFF
--- a/pkg/github/github_client.go
+++ b/pkg/github/github_client.go
@@ -17,9 +17,12 @@ limitations under the License.
 package github
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	ghinstallation "github.com/bradleyfalzon/ghinstallation/v2"
@@ -80,20 +83,25 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
 	return NewGithubClient(token.GetToken()), nil
 }
 
-func (c *GithubClient) GetReference(owner, repository, branch string) (*github.Reference, error) {
+func (c *GithubClient) referenceExist(owner, repository, branch string) (bool, error) {
+	_, resp, err := c.client.Git.GetRef(c.ctx, owner, repository, "refs/heads/"+branch)
+	if err == nil {
+		return true, nil
+	}
+
+	if resp.StatusCode == 404 {
+		return false, nil
+	}
+	return false, err
+}
+
+func (c *GithubClient) getReference(owner, repository, branch string) (*github.Reference, error) {
 	ref, _, err := c.client.Git.GetRef(c.ctx, owner, repository, "refs/heads/"+branch)
 	return ref, err
 }
 
-func (c *GithubClient) GetOrCreateBranchReference(owner, repository, branch, baseBranch string) (*github.Reference, error) {
-	ref, resp, err := c.client.Git.GetRef(c.ctx, owner, repository, "refs/heads/"+branch)
-	if err == nil {
-		return ref, nil
-	} else if resp.StatusCode != 404 {
-		return nil, err
-	}
-
-	baseBranchRef, err := c.GetReference(owner, repository, baseBranch)
+func (c *GithubClient) createReference(owner, repository, branch, baseBranch string) (*github.Reference, error) {
+	baseBranchRef, err := c.getReference(owner, repository, baseBranch)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/github/github_client.go
```diff
@@ -80,32 +83,69 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
...
+func (c *GithubClient) isReferenceExist(owner, repository, branch string) (bool, error) {
+	_, resp, err := c.client.Git.GetRef(c.ctx, owner, repository, "refs/heads/"+branch)
+	if err == nil {
+		return true, nil
+	}
+
+	if resp.StatusCode == 404 {
+		return false, nil
+	}
+	return false, err
```
gabemontero, line 95: while I did somewhat, without backtracking all the error propagation @mmorhun, so perhaps what I'm about to mention is "handled", but from what I saw, I *think* retry will occur if a non-404 error occurs (of course correct me if I'm wrong).

if so, there are some other errors besides 404, like permission denied, which are permanent failures which most likely will not go away anytime soon, and some like throttling which might go away relatively soon, and warrant a retry

and hence it could make sense to broaden our http resp code analysis somewhat 

http internal errors, client/server timeouts, connection reset, unexpected eof, etc.

pkg/github/github_client.go
```diff
@@ -80,32 +83,69 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
...
+func (c *GithubClient) isFilesUpToDate(owner, repository, branch string, files []File) (bool, error) {
+	for _, file := range files {
+		opts := &github.RepositoryContentGetOptions{
+			Ref: "refs/heads/" + branch,
+		}
+
+		fileContentReader, resp, err := c.client.Repositories.DownloadContents(c.ctx, owner, repository, file.Name, opts)
+		if err != nil {
+			// It's not clear when it returns 404 or 200 with the error message. Check both.
+			if resp.StatusCode == 404 || strings.Contains(err.Error(), "no file named") {
```
gabemontero, line 130: same sort of potential broadening around which status code's to worry about

gabemontero (commented):
just a thought around potentially more complete http return code analysis with respect to retry vs. aborting @mmorhun

pkg/github/github_client.go
mmorhun: You are right, @gabemontero , the errors must be handled better. We already have a separate issue about that.
In this PR, I am fixing logic of PR creation. Errors handling is untouched.

pkg/github/github_client.go
mmorhun: We can got both responses. That's the only way to handle correctly.

pkg/github/github_client.go
```diff
@@ -80,32 +83,69 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
 	return NewGithubClient(token.GetToken()), nil
 }
 
-func (c *GithubClient) GetReference(owner, repository, branch string) (*github.Reference, error) {
+func (c *GithubClient) isReferenceExist(owner, repository, branch string) (bool, error) {
```
brunoapimentel, line 86: nitpick: I'd call it `doesReferenceExist` or just `referenceExists`.

pkg/github/github_client.go
```diff
@@ -80,32 +83,69 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
...
 	return ref, err
 }
 
-func (c *GithubClient) CreateTree(owner, repository string, baseRef *github.Reference, files []File) (tree *github.Tree, err error) {
+func (c *GithubClient) deleteReference(owner, repository, branch string) error {
+	_, err := c.client.Git.DeleteRef(c.ctx, owner, repository, "refs/heads/"+branch)
+	return err
+}
+
+func (c *GithubClient) isFilesUpToDate(owner, repository, branch string, files []File) (bool, error) {
```
brunoapimentel, line 121: another nitpick: I guess it should be `areFilesUpToDate`. (unless there's some naming schema I'm not aware of)

brunoapimentel (approved):
Did a few tests locally on the new scenarios using the `github_helper_debug_test.go`.
Considering we're improving the test in a following story, this LGMT.

pkg/github/github_client.go
gabemontero: I'm good tracking with the tracking story for handling k8s client non 200 http rc's separately / holistically across the broader component / appstudio even @mmorhun

gabemontero (approved):
/lgtm

PR #54, opened 2022-08-22
Implement PaC GitLab integration

mmorhun:

Note: currently, Application Service fails to download devfile from GitLab. To workaround this, **URL to the devfile** must be explicitly set in the Component CR.

Example:
```yaml
kind: Component
metadata:
  annotations:
    pipelinesascode: "1"
  generation: 1
  name: devfile-sample-go-basic
  namespace: default
spec:
  application: test-application
  componentName: devfile-sample-go-basic
  containerImage: quay.io/username/devfile-sample-go-basic
  resources: {}
  source:
    git:
      url: https://gitlab.com/username/devfile-sample-go-basic
      devfileUrl:  https://gitlab.com/username/devfile-sample-go-basic/-/raw/main/devfile.yaml?inline=false
```

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -49,6 +49,7 @@ import (
 	"github.com/redhat-appstudio/application-service/gitops"
 	gitopsprepare "github.com/redhat-appstudio/application-service/gitops/prepare"
 	"github.com/redhat-appstudio/build-service/pkg/github"
+	"github.com/redhat-appstudio/build-service/pkg/gitlab"
 	tektonapi "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 )
 
@@ -479,26 +480,21 @@ func ConfigureRepositoryForPaC(component appstudiov1alpha1.Component, config map
 		accessToken = strings.TrimSpace(string(config[gitops.GetProviderTokenKey(gitProvider)]))
 	}
 
+	// https://github.com/owner/repository
+	gitSourceUrlParts := strings.Split(component.Spec.Source.GitSource.URL, "/")
+
+	commitMessage := "Appstudio update " + component.Name
+	branch := "appstudio-" + component.Name
+	baseBranch := "main"
+	mrTitle := "Appstudio update " + component.Name
+	mrText := "Pipelines as Code configuration proposal"
+	authorName := "redhat-appstudio"
+	authorEmail := "appstudio@redhat.com"
+
 	switch gitProvider {
 	case "github":
-		// https://github.com/owner/repository
-		gitSourceUrlParts := strings.Split(component.Spec.Source.GitSource.URL, "/")
 		owner := gitSourceUrlParts[3]
-		prData := &github.PaCPullRequestData{
-			Owner:         owner,
-			Repository:    gitSourceUrlParts[4],
-			CommitMessage: "Appstudio update " + component.Name,
-			Branch:        "appstudio-" + component.Name,
-			BaseBranch:    "main",
-			PRTitle:       "Appstudio update " + component.Name,
-			PRText:        "Pipelines as Code configuration proposal",
-			AuthorName:    "redhat-appstudio",
-			AuthorEmail:   "appstudio@redhat.com",
-			Files: []github.File{
-				{Name: ".tekton/" + component.Name + "-" + PipelineRunOnPushFilename, Content: pipelineOnPush},
-				{Name: ".tekton/" + component.Name + "-" + PipelineRunOnPRFilename, Content: pipelineOnPR},
-			},
-		}
+		repository := gitSourceUrlParts[4]
 
 		var ghclient *github.GithubClient
 		if isAppUsed {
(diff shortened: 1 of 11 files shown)


controllers/component_build_controller.go
```diff
@@ -543,8 +549,37 @@ func ConfigureRepositoryForPaC(component appstudiov1alpha1.Component, config map
...
+		}
+
+		gitlabNamespace := gitSourceUrlParts[3]
+		gitlabProjectName := gitSourceUrlParts[4]
+		projectPath := gitlabNamespace + "/" + gitlabProjectName
+
+		err = gitlab.SetupPaCWebhook(glclient, projectPath, webhookTargetUrl, webhookSecret)
+		if err != nil {
+			return "", err
+		}
```
gabemontero, line 564: I'm not concerned about testing for errors when getting the client, but is there any validation of the inputs into `gitlab.SetupPaCWebhook(glclient, projectPath, webhookTargetUrl, webhookSecret)` that make sense (and maybe add unit tests around those?

gabemontero: where we might provide betterr error information to the user than the libraries we are calling?

controllers/component_build_controller.go
```diff
@@ -543,8 +549,37 @@ func ConfigureRepositoryForPaC(component appstudiov1alpha1.Component, config map
...
+			MrTitle:       mrTitle,
+			MrText:        mrText,
+			AuthorName:    authorName,
+			AuthorEmail:   authorEmail,
+			Files: []gitlab.File{
+				{FullPath: ".tekton/" + component.Name + "-" + PipelineRunOnPushFilename, Content: pipelineOnPush},
+				{FullPath: ".tekton/" + component.Name + "-" + PipelineRunOnPRFilename, Content: pipelineOnPR},
+			},
+		}
+		mrUrl, err := gitlab.EnsurePaCMergeRequest(glclient, mrData)
```
gabemontero, line 580: similar validation of input question here

gabemontero (commented):
a couple of minor pre validation before calling dependencies @mmorhun which I are outside the http response code stuff we previously discussed

I do not consider them blockers per se, and full disclosure I did not research the implementations of the github/githlab related libs we are now calling to see if problems inputs are sufficiently described.

pkg/gitlab/gitlab_client.go
```diff
@@ -0,0 +1,220 @@
...
+			return false, nil
+		}
+		if !bytes.Equal(fileContent, file.Content) {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (c *GitlabClient) commitFilesIntoBranch(projectPath, branchName, baseBranch, commitMessage, authorName, authorEmail string, files []File) error {
```
brunoapimentel, line 88: What is the variable `baseBranch` being used for here?

controllers/component_build_controller.go
mmorhun: > is there any validation of the inputs into gitlab.SetupPaCWebhook(glclient, projectPath, webhookTargetUrl, webhookSecret)

`glclient` is created above. The only argument it gets is the access token.
`projectPath` is built from the repository url. If the url is wrong we'll get 404 later.
`webhookTargetUrl` is read from the Openshift cluster
`webhookSecret` is generated by our code

To sum up the above: we might want to have a better validation of user data (token and git url specifically) by querying something from the project just to check if everything works. What do you think?
Anyway, I'd like to leave that implementation for later, if any.

> where we might provide betterr error information to the user than the libraries we are calling?

I am not sure I got the question correctly, but my plan for the future is to have wrapper/custom error and just log/create event for it.

pkg/gitlab/gitlab_client.go
mmorhun: Leftovers after changing the implementations. Will delete that parameter.

controllers/component_build_controller.go
mmorhun: Merge request data struct is filled by the operator based on the component fields (name) and git repository url (user name, repository name). If check for git repository url correctness is added, then we are 100% safe here.

mmorhun: Moreover, if the git repository url is invalid, I think application service will fail first (on detecting devfile), so build service will not start it's work.
However, I'd still have the check as the logic could be changed one day.

brunoapimentel:
Should we add a test scenario that covers the use of the Gitlab client/helper (similar to the ones we currently have covering Github integration)?

Michkov (approved):

controllers/component_build_controller.go
gabemontero: > > is there any validation of the inputs into gitlab.SetupPaCWebhook(glclient, projectPath, webhookTargetUrl, webhookSecret)
> 
> `glclient` is created above. The only argument it gets is the access token. `projectPath` is built from the repository url. If the url is wrong we'll get 404 later. `webhookTargetUrl` is read from the Openshift cluster `webhookSecret` is generated by our code
> 
> To sum up the above: we might want to have a better validation of user data (token and git url specifically) by querying something from the project just to check if everything works. What do you think? Anyway, I'd like to leave that implementation for later, if any.

yep for what it is worth openshift builds does some git ls-remote calls (much less expensive then git clone) to validate git related inputs and catch errors as soon as possible, so there is precedence for such a validation, where that validation is proven useful

I'm fine if you want to leave that for later.  Either add it to your existing jira task around error handling, or open a similar but separate item.

Feel free to mark this item resolved once you've captured things in a Jira.

> 
> > where we might provide betterr error information to the user than the libraries we are calling?
> 
> I am not sure I got the question correctly, but my plan for the future is to have wrapper/custom error and just log/create event for it.

that sounds ^^ exactly around the path I was envisioning

controllers/component_build_controller.go
gabemontero: sounds good ... feel free to resolve based on your updates per the comment thread above

PR #56, opened 2022-08-25
fix: use label-filter for running e2e

psturc:
### Issue description
The current e2e filter does not filter out jvm-build-service tests

Use ginkgo's label-filter to run only [build-service tests](https://github.com/redhat-appstudio/e2e-tests/blob/37b2d326d074417c83853c47847268a29d9e5ced/tests/build/build.go#L43)

### DIFF
--- a/.ci/oci-e2e-build-service.sh
+++ b/.ci/oci-e2e-build-service.sh
@@ -11,7 +11,7 @@ command -v kubectl >/dev/null 2>&1 || { echo "kubectl is not installed. Aborting
 export WORKSPACE BUILD_SERVICE_PR_OWNER BUILD_SERVICE_PR_SHA
 
 WORKSPACE=$(dirname "$(dirname "$(readlink -f "$0")")");
-export TEST_SUITE="build-service-suite"
+export TEST_SUITE_LABEL="build"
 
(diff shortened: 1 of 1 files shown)


gabemontero:
seems ok, but need to make sure that not only the e2e passes, but it includes the tests we want were in fact included

I'll let you post evidence of that @psturc and then can tag this for merge

psturc:
A test flake that happened in [this spec](https://github.com/redhat-appstudio/e2e-tests/blob/37b2d326d074417c83853c47847268a29d9e5ced/tests/build/build.go#L264-L273) and the failure was caused by a conflict in [test repo](https://github.com/redhat-appstudio-qe/devfile-sample-hello-world/commits/main) created by other build-service test running almost at the same time.

However I think we should not worry about it, because that test will be replaced by [PaC build test](https://github.com/redhat-appstudio/e2e-tests/pull/177) which works with unique branches instead of main branch (which means that such conflicts should not occur)
/test build-service-e2e

psturc:
@gabemontero happy to merge it?

see the [log from CI](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_build-service/56/pull-ci-redhat-appstudio-build-service-main-build-service-e2e/1562804452023341056/artifacts/build-service-e2e/build-service-e2e/build-log.txt) - no occurrence of "jvm-build-service", only build-service related tests ran this time

note that the failing `codecov/project` check is not not related to my PR - I guess it's caused by my branch not being up-to-date with `main`

PR #57, opened 2022-08-25
PLNSRVCE-431: Adds annotation to allow skipping the generation of PaC resources

brunoapimentel:
In order to migrate the components in the infra-deployments repository to use HAS Application/Component CRs, we need to avoid having the build-service spamming PRs for each time a new Component is created when a new cluster is bootrstraped.

This PR allows this behavior to be skipped in case an annotation is set in the related Component CR.

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -63,6 +63,8 @@ const (
 
 	PartOfLabelName           = "app.kubernetes.io/part-of"
 	PartOfAppStudioLabelValue = "appstudio"
+
+	skipPacResourceGenerationAnnotation = "skip-pipelines-as-code-resource-generation"
 )
 
 // ComponentBuildReconciler watches AppStudio Component object in order to submit builds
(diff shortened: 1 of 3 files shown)


controllers/component_build_controller.go
```diff
@@ -63,6 +63,8 @@ const (
 
 	PartOfLabelName           = "app.kubernetes.io/part-of"
 	PartOfAppStudioLabelValue = "appstudio"
+
+	skipPacResourceGenerationAnnotation = "skip-pipelines-as-code-resource-generation"
```
gabemontero, line 67:
```suggestion
	skipPacResourceGenerationAnnotation = "com.redhat.appstudio/skip-pipelines-as-code-resource-generation"
```

it is better to qualify labels / annotations with a group designation ... my suggestion reuses the group used for the initial build annotation

controllers/component_build_controller.go
```diff
@@ -216,21 +218,25 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		if mrUrl != "" {
-			mrMessage = fmt.Sprintf("Pipelines as Code configuration merge request: %s", mrUrl)
+		if val, ok := component.Annotations[skipPacResourceGenerationAnnotation]; ok && val == "1" {
+			r.EventRecorder.Event(&component, "Normal", "PipelinesAsCodeConfiguration", "Skipping the generation of .tekton resources")
 		} else {
-			mrMessage = "Pipelines as Code configuration is up to date"
+			bundle := gitopsprepare.PrepareGitopsConfig(ctx, r.NonCachingClient, component).BuildBundle
+			mrUrl, err := ConfigureRepositoryForPaC(component, pacSecret.Data, webhookTargetUrl, webhookSecretString, bundle)
+			if err != nil {
+				log.Error(err, "failed to setup repository for Pipelines as Code")
```
gabemontero, line 227: I've noticed recently in jvm-build-service that with our default level of logging, controller runtime will log the generation of the event in the pod log as well; so you'll end up with essentially duplicate logs with this one ^^ and the event below.

Either in a manual bringup, or investigating the build-service logs collected in the artifacts of the e2e test, you should confirm if that is the case with build-service, and if so, just go with the event here.

@mmorhun FYI

gabemontero (changes requested):
just a coule of minor items @brunoapimentel

controllers/component_build_controller.go
mmorhun: Yes, I can confirm, that events are logged as well and we have duplicates in logs (my mistake before).

PR #61, opened 2022-08-31
PLNSRVCE-642: handle error returned from SearchForDockerfile

tkdchen:
The found devfile could be invalid. This change will propagate the
potential error to Reconcile funtion in order to handle it properly from
the controller level. And the existing code is able to log relative
error message.

Signed-off-by: Chenxiong Qi <cqi@redhat.com>

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -687,7 +687,7 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, o
 	yamlformat, err := yaml.Marshal(pipelineRun)
 	if err != nil {
 		// Should never happen because the function is covered by tests
-		panic(err)
+		return nil, err
 	}
 
 	return yamlformat, nil
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -471,8 +471,14 @@ func (r *ComponentBuildReconciler) EnsurePaCRepository(ctx context.Context, comp
 // ConfigureRepositoryForPaC creates a merge request with initial Pipelines as Code configuration
 // and configures a webhook to notify in-cluster PaC unless application (on the repository side) is used.
 func ConfigureRepositoryForPaC(component appstudiov1alpha1.Component, config map[string][]byte, webhookTargetUrl, webhookSecret, buildBundle string) (prUrl string, err error) {
-	pipelineOnPush := GeneratePipelineRun(component, buildBundle, false)
-	pipelineOnPR := GeneratePipelineRun(component, buildBundle, true)
+	pipelineOnPush, err := GeneratePipelineRun(component, buildBundle, false)
+	if err != nil {
+		return "", err
+	}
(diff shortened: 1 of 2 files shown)


controllers/component_build_controller.go
```diff
@@ -678,7 +690,7 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, o
 		panic(err)
 	}
 
-	return yamlformat
+	return yamlformat, nil
```
mmorhun, line 693: You can return an error above as well, instead of panic.

controllers/component_build_controller.go
```diff
@@ -633,14 +639,20 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, o
 		{Name: "revision", Value: tektonapi.ArrayOrString{Type: "string", StringVal: "{{revision}}"}},
 		{Name: "output-image", Value: tektonapi.ArrayOrString{Type: "string", StringVal: proposedImage}},
 	}
-	if dockerFile, err := devfile.SearchForDockerfile([]byte(component.Status.Devfile)); err == nil && dockerFile != nil {
+
+	dockerFile, err := devfile.SearchForDockerfile([]byte(component.Status.Devfile))
```
mmorhun, line 643: Do we distinguish between not found and other types of errors?

controllers/component_build_controller.go
```diff
@@ -633,14 +639,20 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, o
...
+	if err != nil {
+		return nil, err // devfile could be invalid to cause parser error
```
mmorhun, line 645: Could you elaborate more on this?

tkdchen: Good question. After rethinking, I think this comment may not be accurate to describe the possible error. Any error from `SearchForDockerfile` should cause `GeneratePipelineRun` returns immediately. This also answers your question above.

mmorhun (approved):

Michkov (approved):

controllers/component_build_controller.go
```diff
@@ -633,14 +639,20 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, o
 		{Name: "revision", Value: tektonapi.ArrayOrString{Type: "string", StringVal: "{{revision}}"}},
 		{Name: "output-image", Value: tektonapi.ArrayOrString{Type: "string", StringVal: proposedImage}},
 	}
-	if dockerFile, err := devfile.SearchForDockerfile([]byte(component.Status.Devfile)); err == nil && dockerFile != nil {
+
+	dockerFile, err := devfile.SearchForDockerfile([]byte(component.Status.Devfile))
+	if err != nil {
+		return nil, err
+	}
```
tkdchen, line 646: Tested with an invalid devfile by missing `schemaVersion`. The controller's `Reconcile` method returns earlier at the step of checking if `Status.Devfile` is empty or not. No chance to handle the error returned from this call.

PR #62, opened 2022-09-19
PLNSRVCE-700: use volumeClaimTemplate instead of static PVC

tkdchen:
Signed-off-by: Chenxiong Qi <cqi@redhat.com>


controllers/component_build_controller.go
```diff
@@ -672,9 +667,19 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, o
 			Params: params,
 			Workspaces: []tektonapi.WorkspaceBinding{
 				{
-					Name:                  "workspace",
-					PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{ClaimName: "appstudio"},
-					SubPath:               pipelineName + "-{{revision}}",
+					Name: "workspace",
+					VolumeClaimTemplate: &corev1.PersistentVolumeClaim{
```
mmorhun, line 671: Does it makes sense to have a function that generates the `VolumeClaimTemplate`? Then use the function in application-service and here, so we have consistency on how the volume is created.

tkdchen: Good point :+1:

tkdchen:
Update: this pr includes the changes made by Michal in PR #63.

go.mod
```diff
@@ -18,15 +18,23 @@ require (
 )
 
 require (
-	github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0 // indirect
-	golang.org/x/tools v0.1.10-0.20220218145154-897bd77cd717 // indirect
+	code.gitea.io/sdk/gitea v0.14.0 // indirect
```
mmorhun, line 21: Please group all indirect dependencies under the same require block

controllers/component_build_controller.go
```diff
@@ -618,8 +612,7 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, o
 	labels := map[string]string{
 		ComponentNameLabelName:                  component.Name,
 		"pipelines.appstudio.openshift.io/type": "build",
-		"appstudio.openshift.io/application":    component.Spec.Application,
-		"appstudio.openshift.io/component":      component.Name,
+		ApplicationNameLabelName:                component.Spec.Application,
```
mmorhun, line 615: I'd prefer to have application and component labels one next to the other (not separated by another label).

mmorhun (approved):
LGTM if my remarks fixed

controllers/component_build_controller.go
tkdchen: Both are updated. PTAL.

Michkov (approved):

PR #64, opened 2022-09-26
Draft. Make build service KCP aware

mmorhun:


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
...
+            "request": "launch",
+            "name": "Debug App",
+            "program": "${workspaceFolder}/main.go",
+            "args": [
+                "--api-export-name", "build-service-apiexport",
+                "--zap-log-level", "6",
+                "--zap-devel", "true",
+            ],
+            "env": {
+                "KUBECONFIG":"/home/user/projects/kcp/.kcp/admin.kubeconfig"
```
Michkov, line 15: this is too much specific so the file should not be included in the commit.

Makefile
```diff
@@ -132,7 +135,7 @@ endif
 
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+	$(KUSTOMIZE) build config/crd | kubectl apply -f - || true
```
Michkov, line 138: why failed apply is ignored?

Makefile
```diff
@@ -155,7 +163,7 @@ controller-gen: ## Download controller-gen locally if necessary.
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 .PHONY: kustomize
 kustomize: ## Download kustomize locally if necessary.
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v3.8.7)
+	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v4@v4.5.2)
```
Michkov, line 166: nit: it's probably rewrite commands to use `kubectl apply -k xxx` instead of `kustomize build xxx | kubectl apply -f -`

.vscode/launch.json
mmorhun: ok, will remove it

Makefile
mmorhun: Currently, build service does not bring any new CRDs, so build for empty dir `config/crd` returns nothing. To avoid error for the make target, the ignore was added.
I decided to keep `install` command and `crd` directory to not to break common pattern. But if you think that it's better to remove them - let me know.

Makefile
mmorhun: This makes sense, but let's leave it as is for now.

go.mod
```diff
@@ -4,24 +4,20 @@ go 1.17
 
 require (
 	github.com/go-logr/logr v1.2.3
+	github.com/kcp-dev/kcp/pkg/apis v0.7.0
```
gabemontero, line 7: both containerized kcp in openshift-pipelines/placement-service and CPS stable as I understand it are at v0.8.2

it may not have an impact on your api consumption based on what I see in your yaml, but conversely, it might be good to stay in sync with the test envs

main.go
```diff
@@ -219,3 +237,137 @@ func getCacheFunc() (cache.NewCacheFunc, error) {
...
+				if version.Version == apisv1alpha1.SchemeGroupVersion.Version {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
+func ensureRequiredAPIGroupsAndResourcesExist(restConfig *rest.Config) {
```
gabemontero, line 309: I will probably borrow this for jvm-build-service :-)

main.go
```diff
@@ -219,3 +237,137 @@ func getCacheFunc() (cache.NewCacheFunc, error) {
...
+					// Required group exists
+					// Check for required resources of the group
+				NextResource:
+					for _, requiredGroupResource := range requiredGroupResources {
+						for _, apiResource := range apiResources {
+							groupName := strings.Split(apiResource.GroupVersion, "/")[0]
+							if groupName == requiredGroupName {
+								for _, apiResourceInGroup := range apiResource.APIResources {
+									if apiResourceInGroup.Name == requiredGroupResource {
+										continue NextResource
```
gabemontero, line 349: I guess employing labels and continues like this optimizes going through a a lengthy list of groups, or where some groups have a ton of resources, and let's you use ?

Augmenting your map definition to a map of maps could help with tracking status and avoid the jumping around with continue, where you instead remove resources, then groups from `requiredGroupsAndResources`
- populate `requiredGroupsAndResources`as a map[string]map[string]struct{}
- rather than range'ing over `requiredGroupsAndResources`, just range `apiGroups` and `apiResources` returned from `discoveryClient.ServerGroupsAndResources()`
- skip groups whose name when fed into `val, ok := requiredGroupsAndResources[group.Name]` results in `ok` being `false`
- then, with `val` now being a map of `map[string]struct{}`, after you string machinations, if you find a resource who has a key in `val`, then remove it
- when the map[string]struct{} returned from `requiredGroupsAndResources[group.Name]` is at zero length, you have finished that group, so remove the group key from `requiredGroupsAndResources`
- and when `requiredGroupsAndResources` is empty, you have found everything

config/kcp/build-service-apibinding.yaml
```diff
@@ -0,0 +1,41 @@
...
+      exportName: build-service-apiexport
+      path: root:appstudio:build-service
+  permissionClaims:
+  - group: ""
+    resource: configmaps
+    state: Accepted
+  - group: ""
+    resource: secrets
+    state: Accepted
+  - group:
```
gabemontero, line 17: curious that for some of the core k8s groups you have `""` and others you have just a blank

is that by design?

if so, I wonder if the reason why pertains to a problem I currently have where I can get some core k8s items listed in my permission claim, but not others

gabemontero (commented):
A first pass set of comments @mmorhun 

One go.mod suggestion

One golang suggestion our your type discovery. If you choose not to do it based on my description, I may employ it then point you to it for your considerationwhen I copy your use of discovery client in jvm build service to ensure presence of things like tekton.

One question on your permission claims

Looking at this controller's creation of the managers, and only needing predicates, FWIW your KCP aware initialization is simpler / needs less than jvm-build-service's (because of our use of selectors we have to customize the cache function override).

tkdchen:
The first commit message has "Draft." prefix.

main.go
fgiloux: Why not let the container gets into CrashLoopBackOff if an API is not available?

config/kcp/build-service-apibinding.yaml
```diff
@@ -0,0 +1,41 @@
...
+    resource: serviceaccounts
+    state: Accepted
+  - group: tekton.dev
+    identityHash: pipelinesIdentityHash
+    resource: pipelineruns
+    state: Accepted
+  - group: tekton.dev
+    identityHash: pipelinesIdentityHash
+    resource: taskruns
```
Michkov, line 32: taskruns are not provided by pipeline-service

go.mod
mmorhun: @gabemontero good point. I didn't check it, thinking that the correct version was added automatically. Thanks for pointing out.
Thank you, @fgiloux

config/kcp/build-service-apibinding.yaml
mmorhun: There is no special meaning AFAIK. Added `""` for consistency.

main.go
mmorhun: Your approach is good and makes sense (I was thinking about something similar, however I had `bool` instead of `struct{}`), but my point here was to have the clear "interface". It's more friendly to look and at:
```go
"tekton.dev": {
	"pipelineruns",
	"taskruns",
},
"appstudio.redhat.com": {
	"components",
	"applications",
},
```
than have unclear values for each of the resources.
So, I decided to sacrifice a bit code below, but keep the list clear.
Anyway, thank you @gabemontero for the suggestion. If you have strong opinion that we need to go that way, please let me know and I'll rework.

config/kcp/build-service-apibinding.yaml
```diff
@@ -0,0 +1,41 @@
...
+    resource: namespaces
+    state: Accepted
+  - group:
+    resource: events
+    state: Accepted
+  - group:
+    resource: serviceaccounts
+    state: Accepted
+  - group: tekton.dev
+    identityHash: pipelinesIdentityHash
```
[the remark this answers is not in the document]
mmorhun: I thought it would be clear that `pipelinesIdentityHash` is not a real hash. However, I agree with you that it's better to have a note, because one might not read through the yaml and just apply it. Fixing, thanks.

controllers/component_build_controller.go
```diff
@@ -114,6 +116,11 @@ func (r *ComponentBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithValues("ComponentOnboarding", req.NamespacedName)
 
+	if req.ClusterName != "" {
```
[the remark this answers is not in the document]
mmorhun: Field `ClusterName` is KCP specific. Regular `req`does not have it. I though that it was self-explanatory.
If it rises a question, then better to add a comment. Will do, thank for letting me know.

main.go
gabemontero: No as I tried to convey in my last comment, I don't feel super strongly about it.  If you want to keep it as is, that is fine.  Long term, assuming what you are doing even works in KCP (i.e. @Michkov mentioned in stand up yesterday it was not working in KCP), when I borrow this use of discovery client and use it in jvm build service, I most likely will try my `map[string]map[string]struct{}` approach, and assuming it works, we can compare.  Long term, it probably makes sense to have such a utility in a common repo so we can share.  And then we can let the community decide which approach it likes better ;-)

config/kcp/build-service-apibinding.yaml
gabemontero: consistency sounds good to me

config/kcp/build-service-apibinding.yaml
mmorhun: I disabled the controller for the first iteration.

Michkov (approved):
Tested via https://github.com/redhat-appstudio/infra-deployments/pull/775 and I am able to integrate. Creation of Component creates a pipelineRun.

PR #66, opened 2022-10-06
Make Secrets and ConfigMaps visible to the operator

mmorhun:
Do not use `nonCachingClient` any more, instead use regular client with `ClientDisableCacheFor` option that's set to `Secret`s and `ConfigMap`s.
This change was made because regular non-caching client didn't work on KCP. As a consequence, build service operator didn't see build bundle ConfigMap and credentials Secrets, so Component image build was failing.


Michkov (approved):

mmorhun:
@johnbieren there is no error message AFAIK. The client cannot locate the objects, so 404 returned even if the object really exists.

PR #71, opened 2022-10-25
Make PaC Webhook work on KCP

mmorhun:

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -23,6 +23,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


controllers/component_build_controller.go
```diff
@@ -173,9 +175,20 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		log.Info("Pipelines as Code enabled")
 
 		// Obtain Pipelines as Code callback URL
-		webhookTargetUrl, err := r.getPaCRoutePublicUrl(ctx)
-		if err != nil {
-			return ctrl.Result{}, err
+		webhookTargetUrl := os.Getenv(pipelinesAsCodeRouteEnvVar)
+		if webhookTargetUrl == "" {
+			// The env variable is not set
+			if req.ClusterName != "" {
```
gabemontero, line 181:
```suggestion
			if req.ClusterName == "" {
```

if the string is empty, then it is not set, right?

gabemontero (changes requested):

controllers/component_build_controller.go
mmorhun: The comment is related to the if above: `if webhookTargetUrl == "" ` that happens after an attempt to read the env variable.

mmorhun: I'll add an empty line to separate

Michkov (approved):

controllers/component_build_controller.go
gabemontero: ah ok I see making this resolved

gabemontero (approved):
/lgtm

PR #75, opened 2022-11-08
Change expected location of PaC secret

mmorhun:
Expect PaC configuration secret (`pipelines-as-code-secret`) in the same namespace as component. If the secret is missing locally, fallback to global PaC configuration in `build-service` namespace (in case of KCP, in the operator deployment workspace).

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -17,7 +17,6 @@ limitations under the License.
 package controllers
 
 import (
-	"bytes"
 	"context"
 	"crypto/rand"
 	"encoding/hex"
@@ -69,6 +68,7 @@ const (
 	pipelinesAsCodeRouteEnvVar = "PAC_WEBHOOK_URL"
 
 	buildPipelineServiceAccountName = "pipeline"
+	buildServiceNamespaceName       = "build-service"
 	defaultPipelineName             = "docker-build"
 
 	PartOfLabelName           = "app.kubernetes.io/part-of"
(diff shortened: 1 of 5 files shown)


Michkov (approved):

controllers/component_build_controller.go
```diff
@@ -190,10 +190,44 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		localPaCSecretKey := types.NamespacedName{Namespace: component.Namespace, Name: gitopsprepare.PipelinesAsCodeSecretName}
+		if err := r.Client.Get(ctx, localPaCSecretKey, &pacSecret); err != nil {
+			if !errors.IsNotFound(err) {
+				log.Error(err, fmt.Sprintf("failed to get Pipelines as Code secret in %s namespace", component.Namespace))
+				r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
+				return ctrl.Result{}, err
+			}
+
+			// Fallback to the global configuration
+			if err := r.LocalClient.Get(ctx, types.NamespacedName{Namespace: buildServiceNamespaceName, Name: gitopsprepare.PipelinesAsCodeSecretName}, &pacSecret); err != nil {
```
Michkov, line 202: would it be possible to skip the `Namespace` so it would be created in namespace where the operator is running?

mmorhun: No, the `namespace` field is mandatory because it's part of object identifier.

PR #84, opened 2022-12-02
Implement ability to customize build piepline selection logic

mmorhun:

#### What does this PR do

This PR introduces ability to define build pipeline selection logic for components.

Build pipeline selection logic might be defined in CRs of `BuildPipelineSelector` type.
The operator analyzes CRs in the following order:
1. CR in the Application namespace that has name the same as the Application (see `Component.Spec.Application`).
2. CR named `build-pipeline-selector` in the Application namespace
3. CR named `build-pipeline-selector` in the `build-service` namespace

If no matches found or no `BuildPipelineSelector` CRs exists, then `docker-build` pipeline from default bundle is used.

#### How to test it

1. Create namespace `test`
4. Create `BuildPipelineSelector` CR named `build-pipeline-selector` in `test` namespace:
```yaml
kind: BuildPipelineSelector
metadata:
  name: build-pipeline-selector
spec:
  selectors:
    - name: Java
      pipelineRef:
        name: java-builder
        bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-java-builder:be25eed597e24e4b8404bacae3cd493b33a36226
     pipelineParams:
      - name: rebuild
        value: "true" 
     when:
        language: java
    - name: NodeJS
      pipelineRef:
        name: nodejs-builder
        bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-nodejs-builder:be25eed597e24e4b8404bacae3cd493b33a36226
      when:
        language: nodejs,node
    - name: Docker build
      pipelineRef:
        name: docker-build
        bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:be25eed597e24e4b8404bacae3cd493b33a36226
      pipelineParams:
        - name: rebuild
          value: "true"
      when:
        dockerfile: true
    - name: Fallback
      pipelineRef:
        name: noop
        bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:be25eed597e24e4b8404bacae3cd493b33a36226
```
5. Create application:
```yaml
kind: Application
metadata:
  name: java-application
  namespace: test
spec:
  description: Java SpringBoot basic example
  displayName: java-app-test
```
6. Fork [Java Spring Boot sample](https://github.com/devfile-samples/devfile-sample-java-springboot-basic)
7. Create `redhat-appstudio-registry-pull-secret` secret with docker image repository credentials (to push the built image into)
8. Create component (replace placeholders with your values):
```yaml
kind: Component
metadata:
  name: java-component
  namespace: test
spec:
  componentName: java-component
  application: java-application
  source:
    git:
      url: https://github.com/<username>/devfile-sample-java-springboot-basic
  containerImage: docker.io/<username>/java-component:tag
```
9. Check `build-service` operator logs and ensure that `java-builder` pipeline from the `quay.io/redhat-appstudio-tekton-catalog/pipeline-java-builder:be25eed597e24e4b8404bacae3cd493b33a36226` bundle was used. Also `rebuild` build parameter must be set to `true`.


api/v1alpha1/initialbuildpipelineselector_types.go
```diff
@@ -0,0 +1,145 @@
...
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// InitialBuildPipelineSelector is the Schema for the initialbuildpipelineselectors API
+type InitialBuildPipelineSelector struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   InitialBuildPipelineSelectorSpec   `json:"spec,omitempty"`
+	Status InitialBuildPipelineSelectorStatus `json:"status,omitempty"`
```
Michkov, line 131: not sure if we will need status, it's basically static object.

config/crd/patches/cainjection_in_initialbuildpipelineselectors.yaml
```diff
@@ -0,0 +1,7 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
Michkov, line 6: includes dependency on cert-manager which is not currently installed by infra-deploymets. Also the patch is not referenced.

controllers/component_build_controller.go
```diff
@@ -718,6 +721,80 @@ func GeneratePipelineRun(component appstudiov1alpha1.Component, bundle string, p
...
+			}
+			// The config is not found, try the next one in the hierarchy
+		} else {
+			pipelineSelectors = append(pipelineSelectors, *pipelineSelector)
+		}
+	}
+
+	if len(pipelineSelectors) == 0 {
+		// Initial build pipeline selection logic is not defined.
+		// Create default configuration.
```
Michkov, line 752: do we need to handle this? if we just skip the if then it will end-up with deafult fallback bundle (which would happen also in this case)

config/crd/kustomization.yaml
```diff
@@ -2,4 +2,5 @@
 # since it depends on service name and namespace that are out of this kustomize package.
 # It should be run by config/default
 resources:
+- bases/appstudio.redhat.com.redhat.com_initialbuildpipelineselectors.yaml
```
Michkov, line 5: file not included.

config/crd/patches/webhook_in_initialbuildpipelineselectors.yaml
```diff
@@ -0,0 +1,16 @@
+# The following patch enables a conversion webhook for the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
Michkov, line 3: could be included directly in the crd file?

api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1alpha1 contains API Schema definitions for the appstudio.redhat.com v1alpha1 API group
+//+kubebuilder:object:generate=true
+//+groupName=appstudio.redhat.com.redhat.com
```
Michkov, line 19: twice redhat.com (everywhere)

api/v1alpha1/initialbuildpipelineselector_types.go
```diff
@@ -0,0 +1,145 @@
...
+	// The value to compare with is taken from devfile.metadata.projectType field.
+	// +kubebuilder:validation:Optional
+	ProjectType string `json:"projectType,omitempty"`
+
+	// Defines if a Dockerfile should be present in the component, e.g. 'true'.
+	// The value to compare with is taken from devfile components of image type.
+	// +kubebuilder:validation:Optional
+	DockerfileRequired BoolString `json:"dockerfile,omitempty"`
+
+	// Defines names of component to match, e.g. 'my-component'.
```
tkdchen, line 69: Should it be `name` rather than `names`?

pkg/pipeline-selector/pipeline-selector.go
```diff
@@ -0,0 +1,152 @@
...
+		if conditionsMatchComponentParameters(&pipelineSelector.WhenConditions, selectionParameters) {
+			return &pipelineSelector.PipelineRef, pipelineSelector.PipelineParams
+		}
+	}
+	return nil, nil
+}
+
+// conditionsMatchComponentParameters evaluates given pipeline selector against component parameters.
+// In other words, checks if given pipeline can build the component (according to what the pipeline conditions say).
+func conditionsMatchComponentParameters(pipeline, component *buildappstudiov1alpha1.WhenCondition) bool {
```
tkdchen, line 87: Would it be good to specify type for argument `pipeline`?

mmorhun: Both parameters have the same type

api/v1alpha1/initialbuildpipelineselector_types.go
mmorhun: No, because the pipeline supports components with the listed **names**. Probably have to improve description.

api/v1alpha1/groupversion_info.go
mmorhun: Thanks!

config/crd/patches/cainjection_in_initialbuildpipelineselectors.yaml
mmorhun: It's auto-generated by operator-sdk. Will clean up later, thanks.

config/crd/patches/webhook_in_initialbuildpipelineselectors.yaml
mmorhun: Auto-generated. We don't need conversion webhook. Will delete it.

controllers/component_build_controller.go
mmorhun: @Michkov, I thought that it would be nice to have explicit configuration in place, so admin can just edit it instead of searching across the docs and creating the object manually. I'm not insist on it, though. So, if you think that it's redundant, I'll delete it.

api/v1alpha1/initialbuildpipelineselector_types.go
mmorhun: I am not sure either. We might need it for reporting some issues with the config... until then we might have it empty or delete completely.

api/v1alpha1/initialbuildpipelineselector_types.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// WhenConditions defines requirements when specified build pipeline must be used.
+// All conditions are connected via AND, whereas cases within any condition connected via OR.
+// Example:
+//
+//	language: java
+//	projectType: spring,quarkus
+//	annotations:
+//	   builder: gradle,maven
```
stuartwdouglas, line 49: So this means you can't match labels with a comma in the value? Should this be `map[string][]string` instead?

controllers/component_build_controller.go
stuartwdouglas: It does seem odd to me to have a global object lazily created as part of a reconcile loop. If you were going to auto create this wouldn't it make more sense to do it at startup?

pkg/pipeline-selector/pipeline-selector.go
```diff
@@ -0,0 +1,154 @@
...
+//
+// The result is true.
+func pipelineMatchesComponentLabels(pipelineLabels, componentLabels map[string]string) bool {
+	for labelName, labelSupportedValues := range pipelineLabels {
+		if componentLabelValue, componentLabelExists := componentLabels[labelName]; componentLabelExists {
+			if !pipelineMatchesComponentCondition(labelSupportedValues, componentLabelValue) {
+				return false
+			}
+		} else {
+			return false
```
stuartwdouglas, line 149: This does not look right? From your example above if `anotherLable: someValue` was the first item in the loop then `componentLabelExists` would return false, even though you are expecting a true result.

stuartwdouglas (changes requested):

config/crd/kustomization.yaml
mmorhun: Fixed

api/v1alpha1/initialbuildpipelineselector_types.go
mmorhun: I was thinking about it. The thing is that in most cases there will be only one value (and I didn't want to force user to create one-item-list all the time), but for some extra cases we might need a list.
Another option is to use `ArrayOrString` but it still doesn't give that easiness in reading and writing (by the user).

mmorhun: The decision is not final, I'm open to discussion

api/v1alpha1/initialbuildpipelineselector_types.go
mmorhun: AFAIK, [it's not allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) to have comma in the label/annotation key/value. So, shouldn't be a problem.
I was thinking about string array (explained in [one of previous comments](https://github.com/redhat-appstudio/build-service/pull/84#discussion_r1039714775)).

controllers/component_build_controller.go
mmorhun: I'm not sure if we need to create it at all... What do you thinks?

pkg/pipeline-selector/pipeline-selector.go
mmorhun: `anotherLable: someValue` will never be in the top loop because it's label on the component, but we iterate over labels on the pipeline condition. In other words, the component must match all the labels on the pipeline condition, but it can have others, unrelated to the pipeline, labels as well.

stuartwdouglas: Ah, I gave a bad example, but it you have any unrelated label on the pipeline then it will fail to match, which I am not sure is what you want.

E.g. if the pipeline had `unrelated: someValue` then this will cause the match to fail, which basically means that adding any unrelated labels/annotations to a pipeline can change the builder that is selected. Is this really the logic that you are after as this does not really feel right?

stuartwdouglas: Ah, thanks for the example.

stuartwdouglas (approved):

api/v1alpha1/initialbuildpipelineselector_types.go
```diff
@@ -0,0 +1,136 @@
...
+
+	tektonapi "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+)
+
+// Defines boolean value in a string to have not set as well.
+// +kubebuilder:validation:Enum=true;false;True;False;1;0
+type BoolString string
+
+const (
+	Bool_true  BoolString = "true"
```
gabemontero, line 30: per our slack exchange, I thought this was giving you problems @mmorhun on kustomize

did you sort that out, or is simply using string per my suggestion a change you still plan on making

pkg/pipeline-selector/pipeline-selector.go
```diff
@@ -0,0 +1,152 @@
...
+// to find the build pipeline for the component.
+// The first match is returned.
+func SelectPipelineForComponent(component *appstudiov1alpha1.Component, selectors []buildappstudiov1alpha1.InitialBuildPipelineSelector) (*tektonapi.PipelineRef, []tektonapi.Param, error) {
+	selectionParameters, err := getPipelineSelectionParametersForComponent(component)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	for _, selector := range selectors {
+		if buildPipelineRef, buildPipelineAdditionalParams := findMatchingPipeline(selectionParameters, &selector); buildPipelineRef != nil {
```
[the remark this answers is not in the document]
gabemontero: yeah since you only read `selector` in `findMatcingPipeline` at the moment this code scanning complaint is a no-op, but it would be cleaner for use `for i, selector := range selectors` and then use `&selector[i]`

api/v1alpha1/initialbuildpipelineselector_types.go
mmorhun: I decided to go another way and use a pointer to bool. Then we have `true`, `false` and `nil`.

PROJECT
```diff
@@ -6,4 +6,14 @@ plugins:
...
 projectName: build-service
 repo: github.com/redhat-appstudio/build-service
+resources:
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: redhat.com
+  group: appstudio.redhat.com
+  kind: InitialBuildPipelineSelector
```
Michkov, line 16: could be just BuildPipelineSelector

mmorhun: Renamed

controllers/component_build_controller.go
```diff
@@ -103,6 +106,7 @@ func (r *ComponentBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/status,verbs=get;list;watch
+//+kubebuilder:rbac:groups="appstudio.redhat.com",resources=buildpipelineselector,verbs=get;list;watch;create;update;patch
```
Michkov, line 109: buildpipelineselector -> buildpipelineselectors

mmorhun: Indeed, thanks

api/v1alpha1/buildpipelineselector_types.go
```diff
@@ -0,0 +1,127 @@
...
+	// Defines type of project of the component to match, e.g. 'quarkus'.
+	// The value to compare with is taken from devfile.metadata.projectType field.
+	// +kubebuilder:validation:Optional
+	ProjectType string `json:"projectType,omitempty"`
+
+	// Defines if a Dockerfile should be present in the component.
+	// Note, unset (nil) value is not the same as false (unset means skip the dockerfile check).
+	// The value to compare with is taken from devfile components of image type.
+	// +kubebuilder:validation:Optional
+	DockerfileRequired *bool `json:"dockerfile,omitempty"`
```
gabemontero, line 51: yep this is the pattern used in k8s ... +1 per our earlier conversations

pkg/pipeline-selector/pipeline_selector.go
```diff
@@ -0,0 +1,162 @@
...
+			return &pipelineSelector.PipelineRef, pipelineParams
+		}
+	}
+	return nil, nil
+}
+
+// pipelineConditionsMatchComponentParameters evaluates given pipeline selector against component parameters.
+// In other words, checks if given pipeline can build the component (according to what the pipeline conditions say).
+func pipelineConditionsMatchComponentParameters(pipeline, component *buildappstudiov1alpha1.WhenCondition) bool {
+	if pipeline.Language != "" && !pipelineMatchesComponentCondition(pipeline.Language, component.Language) {
```
gabemontero, line 98: switch {
case ...
}

is generally the preferred form vs. if's for what you have here, but I'll still apply the approve label and let you decide on your own if you want to change

gabemontero (approved):
a minor suggestion but will still approve and let @mmorhun decide what to do with it

tkdchen (approved):

PR #94, opened 2023-01-04
PLNSRVCE-911: set workspace with PaC github secret

Michkov:
Added workspace binding for 'git-auth' workspace, set to "{{ git_auth_secret }}" which is replaced by PipelinesAsCode.

Also Workspaces binding is set only when workspace is defined in pipeline.

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -674,6 +674,8 @@ func generatePaCPipelineRunForComponent(component *appstudiov1alpha1.Component, 
 
 	params = mergeAndSortTektonParams(params, additionalPipelineParams)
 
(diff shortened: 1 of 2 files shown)


mmorhun (approved):

tkdchen (approved):

psturc:
this probably requires an update to e2e tests?
```
• [FAILED] [133.103 seconds]
[build-service-suite Build service E2E tests] the component with git source (GitHub) is created [It] triggers a PipelineRun 
```

it times out on detecting the pipelinerun that should be created after component creation.

Michkov:
Pac was not able to detect pipelineRun, but I cannot reproduce it, so trying again.

/retest

PR #96, opened 2023-01-05
Implement initial build provision metrics

mmorhun:
This PR adds two prometheus metrics:
 - Time spent from the component creation till initial build pipeline created.
 - Time spent from the component creation till the component repository is provisioned for Pipelines-as-Code.

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -9,8 +9,13 @@ spec:
   template:
     spec:
       containers:
+      - name: manager
+        args:
+        - "--health-probe-bind-address=:8081"
+        - "--metrics-bind-address=127.0.0.1:8080"
+        - "--leader-elect"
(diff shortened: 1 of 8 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -9,8 +9,13 @@ spec:
   template:
     spec:
       containers:
+      - name: manager
+        args:
+        - "--health-probe-bind-address=:8081"
+        - "--metrics-bind-address=127.0.0.1:8080"
+        - "--leader-elect"
```
Michkov, line 16: should not be ports defined and probes set?

Michkov: ok it's in config/manager/manager.yaml

mmorhun:
After discussion with the team, changed buckets for the histograms, so now they are: `5, 10, 15, 20, 30, 60, 120, 300`

PR #101, opened 2023-01-13
Implement PipelineRun pruner controller

mmorhun:

### DIFF
--- a/controllers/pac_pipelinerun_pruner_controller.go
+++ b/controllers/pac_pipelinerun_pruner_controller.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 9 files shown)


controllers/pac_pipelinerun_pruner_controller.go
```diff
@@ -0,0 +1,126 @@
...
+		return err
+	}
+	componentPipelineRunsSelector := labels.NewSelector().Add(*componentPipelineRunsRequirement)
+	componentPipelineRunsListOptions := client.ListOptions{
+		LabelSelector: componentPipelineRunsSelector,
+		Namespace:     req.Namespace,
+	}
+
+	componentPipelineRunsList := &tektonapi.PipelineRunList{}
+	if err := r.Client.List(ctx, componentPipelineRunsList, &componentPipelineRunsListOptions); err != nil {
```
gabemontero, line 108: I don't think the list prior to the delete provides you any real value (iirc the delete with the list option does a list under the covers for you anyway)

Just do the delete

gabemontero (changes requested):

controllers/pac_pipelinerun_pruner_controller.go
mmorhun: Thank you, Gabe, for the comment.
I do list prior to the delete to log information about cleaning for the Component (not needed or number of pipelines cleaned). Otherwise, we will not have any logs from the controller other than a Component was processed. The list operation is cheap because we read it from the cache (and the cache is customized, so we have only related to AppStudio pipelines there). So, to me the question is whether we need the logs or not.

gabemontero: I personally don't think a log including the deleted items is worth it, even with the fact that you most likely will be hitting the cache on the list.

Code is much more concise with just the delete.

Just log prior to the delete stating you are deleting based on the component / label etc.

Michkov (approved):

PR #103, opened 2023-01-19
Allow PaC onboarding after component creation

mmorhun:
This PR allows to do Pipelines as Code provision for Components after they have been created.
In order to do the PaC provision, add `appstudio.openshift.io/pac-provision=request`annotation to the component. When the provision finished successfully, the annotation will be updated by the operator to `appstudio.openshift.io/pac-provision=done`.

Also, initial build annotation name is changed from `com.redhat.appstudio/component-initial-build-processed` to `appstudio.openshift.io/component-initial-build` which is a **breaking change**.
To submit another initial build just delete the annotation (works only if `appstudio.openshift.io/pac-provision` annotation is not set).

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -62,11 +62,14 @@ import (
 )
 
 const (
-	InitialBuildAnnotationName = "com.redhat.appstudio/component-initial-build-processed"
-	ApplicationNameLabelName   = "appstudio.openshift.io/application"
-	ComponentNameLabelName     = "appstudio.openshift.io/component"
-	PartOfLabelName            = "app.kubernetes.io/part-of"
-	PartOfAppStudioLabelValue  = "appstudio"
+	InitialBuildAnnotationName           = "appstudio.openshift.io/component-initial-build"
+	PaCProvisionAnnotationName           = "appstudio.openshift.io/pac-provision"
+	PaCProvisionRequestedAnnotationValue = "request"
+	PaCProvisionDoneAnnotationValue      = "done"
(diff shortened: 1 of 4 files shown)


controllers/component_build_controller.go
```diff
@@ -194,177 +195,207 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-		pacSecret := corev1.Secret{}
-		pacSecretKey := types.NamespacedName{Namespace: component.Namespace, Name: gitopsprepare.PipelinesAsCodeSecretName}
-		if err := r.Client.Get(ctx, pacSecretKey, &pacSecret); err != nil {
-			if !errors.IsNotFound(err) {
-				log.Error(err, fmt.Sprintf("failed to get Pipelines as Code secret in %s namespace", component.Namespace))
-				r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
-				return ctrl.Result{}, err
-			}
+	// Set initial build annotation to prevent next builds
+	component.Annotations[InitialBuildAnnotationName] = "processed"
```
psturc, line 260: should the annotation value be a constant as well?

mmorhun: It could, but the value are ignored for now and only presence of the annotation matters.

mmorhun:
@psturc thank you for the detailed investigation. I have found and fixed the issue.

psturc (approved):

psturc (approved):
all tests passed

Michkov (approved):

PR #108, opened 2023-02-01
STONEBLD-679: add pagination for github.listinstallation

Michkov:
By default list installation returns only 30 records. Adding pagination so it will go via all records.

### DIFF
--- a/pkg/github/github_client.go
+++ b/pkg/github/github_client.go
@@ -62,16 +62,32 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
 	if err != nil {
 		return nil, err
 	}
-	installations, _, err := client.Apps.ListInstallations(context.Background(), &github.ListOptions{})
-	if err != nil {
-		return nil, err
+
+	opt := &github.RepositoryListByOrgOptions{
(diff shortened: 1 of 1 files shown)


pkg/github/github_client.go
```diff
@@ -62,16 +62,31 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
...
-		if val.GetAccount().GetLogin() == owner {
-			installID = val.GetID()
+	for {
+		installations, resp, err := client.Apps.ListInstallations(context.Background(), &opt.ListOptions)
+		if err != nil {
+			return nil, err
 		}
+		for _, val := range installations {
+			if val.GetAccount().GetLogin() == owner {
+				installID = val.GetID()
```
mmorhun, line 78: We can stop looping as the value found

pkg/github/github_client.go
```diff
@@ -62,16 +62,31 @@ func newGithubClientByApp(appId int64, privateKeyPem []byte, owner string) (*Git
...
+			}
+		}
+		if resp.NextPage == 0 || installID == 0 {
```
mmorhun, line 81: I think the condition is wrong

mmorhun (approved):

PR #114, opened 2023-02-08
Implement PaC pruning on Component deletion

mmorhun:

This PR:
 - Adds functionality that submits a PR into Components repository when it gets deleted. The PR deletes Pipelines as Code definitions related to the Component. Also removes repository webhook if used. Works for both supported git providers: GitHub and GitLab.
 - Introduces a new finalizer `pac.component.appstudio.openshift.io/finalizer` in order to do Pipelines as Code clean up on Component deletion.
 - Splits Pipelines as Code and initial build logic into separate files

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -18,25 +18,10 @@ package controllers
 
 import (
 	"context"
-	"crypto/rand"
-	"encoding/hex"
 	"fmt"
-	"net/url"
-	"os"
-	"path/filepath"
-	"sort"
-	"strconv"
-	"strings"
-	"time"
 
-	"github.com/google/go-containerregistry/pkg/authn"
-	routev1 "github.com/openshift/api/route/v1"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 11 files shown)


controllers/component_build_controller.go
```diff
@@ -226,28 +209,43 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		}
+
+		// Update PaC annotation
 		if len(component.Annotations) == 0 {
 			component.Annotations = make(map[string]string)
 		}
 		component.Annotations[PaCProvisionAnnotationName] = pacAnnotationValue
+
+		// Add finalizer to clean up Pipelines as Code configuration on component deletion
+		if component.ObjectMeta.DeletionTimestamp.IsZero() {
```
tkdchen, line 242: Should be `! component.ObjectMeta.DeletionTimestamp.IsZero` here?

tkdchen, line 242: Another thought, is it worth to make a separate function to examine whether a resource is being deleted? Something like:
```go
func isResourceBeingDeleted() bool {
...
}
if isResourceBeingDeleted(component) {

}
```

mmorhun: It's right, we add finalizer only if deletion has not been requested

mmorhun: We can do that, but the construction above is a commonly used pattern

controllers/component_build_controller.go
```diff
@@ -203,6 +169,25 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+
+		// Delete Pipelines as Code provision finalizer
+		controllerutil.RemoveFinalizer(&component, PaCProvisionFinalizer)
+		if err := r.Client.Update(ctx, &component); err != nil {
+			return ctrl.Result{}, err
+		}
+		log.Info("PaC finalizer removed")
+
+		// Try to clean up Pipelines as Code configuration
+		r.UndoPaCProvisionForComponent(ctx, &component)
```
Michkov, line 186: why to try to UndoPaCProvisionForComponent on nonPaC component?

mmorhun: good catch. Fixed.

psturc:
I can confirm that the CI failures in this PR are related to the issues with CI we were hitting for past couple of days. I've tried to run the build tests locally against cluster-bot cluster with stonesoup that uses this branch of build-service and didn't hit any issues.

Michkov (approved):

PR #137, opened 2023-03-15
STONEBLD-818:Intergate Image Controller operator

mmorhun:
This PR integrates functionality provided by [Image Controller operator](https://github.com/redhat-appstudio/image-controller) into Build Service.

Generated by Image Controller operator image repository (it's data) is set in `image.redhat.com/image` annotation of the component. 
Build Service uses value from `component.Spec.ContainerImage`, but if it's not set, but the Image controller annotation is present, Build Service will use the image from `image.redhat.com/image`.

Image repository provided directly by the user is supported on the level it was supported before: the user should create a docker secret and link it to the `pipeline` service account.

Under the hood, Build Service adds/removes auto generated image registry secret to/from `pipeline` Service Account. To achieve clean up, a new finalizer `image-registry-secret-sa-link.component.appstudio.openshift.io/finalizer` is added.

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -20,8 +20,10 @@ import (
 	"context"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -41,7 +43,8 @@ import (
 const (
 	InitialBuildAnnotationName = "appstudio.openshift.io/component-initial-build"
 
-	PaCProvisionFinalizer = "pac.component.appstudio.openshift.io/finalizer"
+	PaCProvisionFinalizer            = "pac.component.appstudio.openshift.io/finalizer"
+	ImageRegistrySecretLinkFinalizer = "image-registry-secret-sa-link.component.appstudio.openshift.io/finalizer"
 
 	PaCProvisionAnnotationName             = "appstudio.openshift.io/pac-provision"
 	PaCProvisionRequestedAnnotationValue   = "request"
@@ -53,6 +56,9 @@ const (
 	ComponentNameLabelName    = "appstudio.openshift.io/component"
 	PartOfLabelName           = "app.kubernetes.io/part-of"
 	PartOfAppStudioLabelValue = "appstudio"
+
+	ImageRepoAnnotationName         = "image.redhat.com/image"
+	ImageRepoGenerateAnnotationName = "image.redhat.com/generate"
 
 	buildServiceNamespaceName         = "build-service"
 	buildPipelineSelectorResourceName = "build-pipeline-selector"
@@ -175,6 +181,21 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if !component.ObjectMeta.DeletionTimestamp.IsZero() {
 		// Deletion of the component is requested
 
+		if controllerutil.ContainsFinalizer(&component, ImageRegistrySecretLinkFinalizer) {
+			if err := r.removeImageRegistrySecretLink(ctx, &component); err != nil {
+				return ctrl.Result{}, err
+			}
+
+			controllerutil.RemoveFinalizer(&component, PaCProvisionFinalizer)
+			if err := r.Client.Update(ctx, &component); err != nil {
+				return ctrl.Result{}, err
+			}
+			log.Info("image secret link finalizer removed")
+
+			// Requeue to process other finalizers
+			return ctrl.Result{Requeue: true}, nil
+		}
+
 		if controllerutil.ContainsFinalizer(&component, PaCProvisionFinalizer) {
 			// In order not to block the deletion of the Component delete finalizer
 			// and then try to do clean up ignoring errors.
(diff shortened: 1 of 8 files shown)


tkdchen:
PR description is informative :+1:  Can we move those to somewhere else, e.g. README.md or even the commit message?

The description involves how (the implementation details), but it would be also great if it has a general description of what the integration does and an overview of what has to be implemented so that integration can work.

controllers/component_build_controller.go
```diff
@@ -201,9 +276,61 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
+	// Ensure component image repository credentials secret is linked to the pipeline service account
```
Michkov, line 279: Do we want to run this for each component change? How about limit to only when `image.redhat.com/image` changed?

controllers/component_build_controller_common.go
```diff
@@ -0,0 +1,263 @@
...
+	err := r.Client.Get(ctx, types.NamespacedName{Name: imageRepoSecretName, Namespace: serviceAccount.Namespace}, &dockerSecret)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Secret %s is missing", imageRepoSecretName))
+		return false, boerrors.NewBuildOpError(boerrors.EComponentImageRegistrySecretMissing, err)
+	}
+	if dockerSecret.Annotations == nil {
+		dockerSecret.Annotations = map[string]string{}
+	}
+	// We can always use 0 index because the component uses only one image repository
+	if dockerSecret.Annotations["tekton.dev/docker-0"] != expectedTektonAnnotationValue {
```
Michkov, line 123: this annotation should not be used -> it's only for [basic-auth](https://tekton.dev/docs/pipelines/auth/#configuring-basic-auth-authentication-for-docker) -> expects username and  password, image controller provides dockerauth file directly.

controllers/component_build_controller.go
```diff
@@ -201,6 +223,16 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
+	if _, exists := component.Annotations[ImageRepoGenerateAnnotationName]; exists {
+		// Image repository per component strategy is used.
+		// Wait for the image repository to be created.
+		if _, exists := component.Annotations[ImageRepoAnnotationName]; !exists {
+			log.Info("Waiting for component image quay repository to be created")
+			// Do not requeue as after the annotation update a new event will trigger a new reconcile
+			return ctrl.Result{}, nil
+		}
+	}
```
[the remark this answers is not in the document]
mmorhun: That should be done by HAC UI.

controllers/component_build_controller_initial_build.go
```diff
@@ -140,7 +186,12 @@ func generateInitialPipelineRunForComponent(component *appstudiov1alpha1.Compone
 	if component.Spec.Source.GitSource != nil && component.Spec.Source.GitSource.Revision != "" {
 		revision = component.Spec.Source.GitSource.Revision
 	}
-	image := fmt.Sprintf("%s:initial-build-%s-%d", getContainerImageRepository(component.Spec.ContainerImage), getRandomString(5), timestamp)
+
+	imageRepo, err := getImageRepositoryForComponent(component)
+	if err != nil {
+		return nil, err
+	}
+	image := fmt.Sprintf("%s:build-%s-%d", imageRepo, getRandomString(5), timestamp)
```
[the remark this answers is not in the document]
mmorhun: It cannot be, because we check it at the beginning of the reconcile.

controllers/component_build_controller.go
```diff
@@ -54,6 +57,10 @@ const (
 	PartOfLabelName           = "app.kubernetes.io/part-of"
 	PartOfAppStudioLabelValue = "appstudio"
 
+	ImageRepoAnnotationName         = "image.redhat.com/image"
+	ImageRepoGenerateAnnotationName = "image.redhat.com/generate"
+	imageRepoUserSecretName         = "redhat-appstudio-registry-pull-secret"
```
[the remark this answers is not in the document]
mmorhun: Yes, but that would require user to edit the SA.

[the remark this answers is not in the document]
mmorhun: I used `redhat-appstudio-registry-pull-secret` secret name because there is no configuration available for such a secret.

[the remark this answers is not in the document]
Michkov: yes, that's how it works, it's not connected with image controller integration.

controllers/component_build_controller.go
mmorhun: > Do we want to run this for each component change? How about limit to only when image.redhat.com/image changed? 

That's a good idea, but implementing that would require a new controller. Taking into account that switching to a new image repository needs a PR (in case of PaC) and PaC provision is done in this controller. So, I am not sure this approach gives us a gain, unfortunately.

> Or maybe skip when finalizer exists?

That would work if we hadn't supported switching of `ContainerImage` . Otherwise we need to unsure the link.
It does no harm if we don't read the secret itself.

controllers/component_build_controller_common.go
mmorhun: ok, that's simplifies stuff.

controllers/component_build_controller.go
```diff
@@ -182,18 +215,38 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 				return ctrl.Result{}, err
 			}
+			if err == nil { // If pipeline service account found, unlink the secret from it
+				imageRepoGenerated, imageRepoSecretName, err := getComponentImageRepoAndSecretNameFromImageAnnotation(&component)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				if imageRepoGenerated != getContainerImageRepository(component.Spec.ContainerImage) {
+					// User defined repository was used
+					imageRepoSecretName = imageRegistryUserSecretName
```
tkdchen: If user does not specify ContainerImage and image-controller generates the image repo, when execute to here, isn't `component.Spec.ContainerImage` an empty string? How could this case happen?

controllers/component_build_controller.go
```diff
@@ -175,6 +214,41 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				log.Error(err, fmt.Sprintf("Failed to read service account %s in namespace %s", buildPipelineServiceAccountName, req.Namespace))
+				return ctrl.Result{}, err
+			}
+			if err == nil { // If pipeline service account found, unlink the secret from it
+				imageRepoGenerated, imageRepoSecretName, err := getComponentImageRepoAndSecretNameFromImageAnnotation(&component)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				if imageRepoGenerated != getContainerImageRepository(component.Spec.ContainerImage) {
+					// User defined repository was used
```
tkdchen, line 230: If user does not specify ContainerImage and image-controller generates the image repo, when execute to here, isn't component.Spec.ContainerImage an empty string? How could this case happen?

tkdchen (approved):

PR #145, opened 2023-04-19
Remove reference to prometheus id

Michkov:
Without the datasource it works on any instance.


tkdchen:
Changes look good to me. Any context of this change?

Michkov:
@tkdchen https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1681883151606789?thread_ts=1681818825.110779&cid=C02CTEB3MMF -> originally we exported dashboard on old stage instance which set this value, but this value does not work on new instances. It works when identificator is removed.

mmorhun (approved):

PR #146, opened 2023-04-19
Propose to type comment ok-to-test to execute PipelineRun

Michkov:
Currently Pipelines as code requires to type /ok-to-test to execute pipelinerun. Propose to user to type the comment.

Testing PR -> https://github.com/Michkov/devfile-sample-go-basic/pull/13

### DIFF
--- a/controllers/git_tekton_resources_renovater.go
+++ b/controllers/git_tekton_resources_renovater.go
@@ -233,7 +233,7 @@ func generateConfigJS(slug string, repositories []renovateRepository) string {
 				branchPrefix: "rhtap/",
 				commitMessageExtra: "",
 				commitMessageTopic: "RHTAP references",
-				prFooter: "",
+				prFooter: "To execute skipped test pipelines write comment ` + "`/ok-to-test`" + `",
 				prBodyColumns: ["Package", "Change", "Notes"],
 				prBodyDefinitions: { "Notes": "{{#if (or (containsString updateType 'minor') (containsString updateType 'major'))}}:warning:[migration](https://github.com/redhat-appstudio/build-definitions/blob/main/task/{{{replace '%stask-' '' packageName}}}/{{{newVersion}}}/MIGRATION.md):warning:{{/if}}" },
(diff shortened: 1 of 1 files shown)


controllers/git_tekton_resources_renovater.go
```diff
@@ -233,7 +233,7 @@ func generateConfigJS(slug string, repositories []renovateRepository) string {
 				branchPrefix: "rhtap/",
 				commitMessageExtra: "",
 				commitMessageTopic: "RHTAP references",
-				prFooter: "",
+				prFooter: "To execute Skipped RHTAP Pipeline write comment ` + "`/ok-to-test`" + `",
```
arewm, line 236: RHTAP shouldn't be used by itself. Instead, we should spell out the whole phrase. If we want to get away from the specific naming of the service, we can also try to generalize it. Maybe something like calling them test pipelines?

RHTAP == Red Hat Trusted Application Pipeline, so this would also read `To execute skipped Red Hat Trusted Application Pipeline pipeline [...]` which I think is unneeded redundancy.

arewm (commented):
The same PR still didn't run the RHTAP pipeline. Is there something that didn't work or is it unskippable?

controllers/git_tekton_resources_renovater.go
Michkov: Updated.

Michkov:
> The same PR still didn't run the RHTAP pipeline. Is there something that didn't work or is it unskippable?

@arewm it's on my testing cluster which is not set properly. You can test it on your PR https://github.com/spork-madness/node-express-hello/pull/6

arewm (approved):
Wording looks better, thanks. Testing works as well (good to have the nudge).

We can continue to iterate on placement, highlighting, and wording as needed to ensure that users can detect and comprehend the message. For example, I don't know if `test pipelines` is most helpful, maybe CI runs or something would be better, but what we have now is a good improvement.

tkdchen:
:+1:

PR #150, opened 2023-05-03
STONEBLD-1157: set tag expiration on pull-request images

Michkov:
Set 5 days expiration for tags created by pull-request pipelineruns

### DIFF
--- a/controllers/component_build_controller_pac.go
+++ b/controllers/component_build_controller_pac.go
@@ -22,10 +22,12 @@ import (
 	"encoding/hex"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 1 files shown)


mmorhun (approved):

controllers/component_build_controller_pac.go
```diff
@@ -869,6 +871,13 @@ func generatePaCPipelineRunForComponent(
 		{Name: "revision", Value: tektonapi.ArrayOrString{Type: "string", StringVal: "{{revision}}"}},
 		{Name: "output-image", Value: tektonapi.ArrayOrString{Type: "string", StringVal: proposedImage}},
 	}
+	if onPull {
+		expiration := os.Getenv(pipelineRunOnPRExpirationEnvVar)
```
psturc, line 875: should we validate the value provided via env var? i.e. if the value is invalid, log that information and use the default one?

Michkov: added

PR #162, opened 2023-06-02
fix: expiration not getting set

tnevrlka:
[STONEBLD-1341](https://issues.redhat.com/browse/STONEBLD-1341)

expiration did not get set if the environment variable was set


Michkov (approved):

tnevrlka:
Seems to be affected by https://github.com/golangci/golangci-lint/issues/3862?

PR #164, opened 2023-06-08
STONEBLD-1187: Use new build annotations

mmorhun:
This PR switches to the new build request system in Build Service operator. 
Now, all interactions with Build Service are supposed to be done via two annotations on the Component:
 - `build.appstudio.openshift.io/request` contains build request. The annotation should be added with the name of the requested action as the annotation value. Clients should only add this annotation. When Build Service finish processing of the requested action, the annotation is deleted. Allowed values are: 
     - `trigger-simple-build` submits a new simple build pipeline. The build could be requested at any time regardless PaC Component configuration.
     - `configure-pac` requests Pipelines-as-Code provision for the Component
     - `unconfigure-pac` requests Pipelines-as-Code clean up for the Component
 - `build.appstudio.openshift.io/status` contains result of requested action and current build state of the Component. Clients should only read this annotation. The value of the annotation is a JSON. An example with all possible fields set:
```json
{
    "simple": {
        "build-start-time": "Mon, 12 Jun 2023 11:23:18 GMT",
        "error-id": "201",
        "error-message": "Specified secret with git credential not found"
    },
    "pac": {
        "state": "enabled",
        "merge-url": "https://github.com/user/repo/pull/1234",
        "error-id": "70",
        "error-message": "GitHub Application is not installed in user repository"
    },
    "message": "unexpected build request: unsupported-build-request" 
}
```
Note, that:
 - error fields appear only if an error happens.
 - `pac.state` values are `enabled`, `disabled` and `error` (error fields are set). If whole `pac` section is missing, PaC state is considered disabled (Missing `pac` section shows that PaC was never requested on this component before, where as `disabled` means that it was provisioned and then removed).

<hr>

Right after Component creation, Build Service tries to perform `trigger-simple-build` action as it was done before.

Operability of old annotations is kept for backward compatibility, but will be removed soon.
Deprecated annotations are:
 - `appstudio.openshift.io/component-initial-build`
 - `appstudio.openshift.io/pac-provision`
 - `appstudio.openshift.io/pac-provision-error`

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -41,6 +43,13 @@ import (
 )
 
 const (
+	BuildRequestAnnotationName                    = "build.appstudio.openshift.io/request"
+	BuildRequestTriggerSimpleBuildAnnotationValue = "trigger-simple-build"
+	BuildRequestConfigurePaCAnnotationValue       = "configure-pac"
+	BuildRequestUnconfigurePaCAnnotationValue     = "unconfigure-pac"
+
+	BuildStatusAnnotationName = "build.appstudio.openshift.io/status"
+
 	InitialBuildAnnotationName = "appstudio.openshift.io/component-initial-build"
 
 	PaCProvisionFinalizer            = "pac.component.appstudio.openshift.io/finalizer"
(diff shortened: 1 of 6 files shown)


controllers/component_build_controller_test.go
```diff
@@ -781,98 +949,552 @@ var _ = Describe("Component initial build controller", func() {
...
-			}, timeout, interval).Should(BeFalse())
 		})
 
-		It("should not block component deletion if PaC definitions removal failed", func() {
-			UndoPaCMergeRequestFunc = func(string, *gp.MergeRequestData) (webUrl string, err error) {
-				return "", fmt.Errorf("failed to create PR")
-			}
-			DeletePaCWebhookFunc = func(string, string) error {
-				return fmt.Errorf("failed to delete webhook")
+		It("should provision PaC definitions after initial build, use simple build while PaC enabled, and be able to switch back to simple build only", func() {
```
tkdchen, line 977: Does it test three cases? Can we split it into three separate cases?

controllers/component_build_controller.go
```diff
@@ -110,6 +119,34 @@ func getProvisionTimeMetricsBuckets() []float64 {
 	return []float64{5, 10, 15, 20, 30, 60, 120, 300}
 }
 
+type BuildStatus struct {
+	Simple *SimpleBuildStatus `json:"simple,omitempty"`
+	PaC    *PaCBuildStatus    `json:"pac,omitempty"`
+	// Shows build methods agnostic messages, e.g. invalid build request.
+	Message string `json:"message,omitempty"`
```
tkdchen, line 126: From the requester point of view, if Message is not empty, it means the request is invalid and cannot be handled by build-service, and requester is able to get the reason provided by build-service from this Message field. Is my understanding correct?

controllers/component_build_controller_test.go
mmorhun: It indeed tests 3 cases, but if we split it into 3 tests, then we'd need to do preparation for each step, which is duplicating the work.

controllers/component_build_controller.go
mmorhun: This field is used to show error for invalid build requests. Otherwise set to `done`.

controllers/component_build_controller.go
```diff
@@ -511,6 +512,39 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				break
+			}
+			// Outdated version of the component, wait more.
+		} else {
+			if errors.IsNotFound(err) {
+				// The component was deleted
+				isComponentInCacheUpToDate = true
+				break
+			}
+			log.Error(err, "failed to get the compoennt", l.Action, l.ActionView)
```
Michkov, line 540: compoennt

mmorhun: thanks, fixed

Michkov (approved):

PR #170, opened 2023-06-15
Fix Codecov excludes

mmorhun:


tnevrlka:
I believe this is a duplicate of #169

mmorhun:
@tnevrlka yes, it's a duplicate

PR #171, opened 2023-06-16
Return error when no pipeline is selected

tkdchen:
[STONEBLD-1190](https://issues.redhat.com//browse/STONEBLD-1190)

### DIFF
--- a/controllers/component_build_controller_common.go
+++ b/controllers/component_build_controller_common.go
@@ -74,11 +74,7 @@ func (r *ComponentBuildReconciler) GetPipelineForComponent(ctx context.Context, 
 		return pipelineRef, pipelineParams, nil
 	}
 
-	// Fallback to the default pipeline
-	return &tektonapi.PipelineRef{
-		Name:   defaultPipelineName,
-		Bundle: defaultPipelineBundle,
-	}, nil, nil
+	return nil, nil, boerrors.NewBuildOpError(boerrors.EBuildPipelineSelectorNotDefined, nil)
 }
 
 func (r *ComponentBuildReconciler) ensurePipelineServiceAccount(ctx context.Context, namespace string) (*corev1.ServiceAccount, error) {
--- a/controllers/component_build_controller_common.go
+++ b/controllers/component_build_controller_common.go
@@ -68,9 +68,10 @@ func (r *ComponentBuildReconciler) GetPipelineForComponent(ctx context.Context, 
 		if err != nil {
 			return nil, nil, err
 		}
-		if pipelineRef != nil {
-			return pipelineRef, pipelineParams, nil
+		if pipelineRef == nil {
+			return nil, nil, boerrors.NewBuildOpError(boerrors.ENoPipelineIsSelected, nil)
 		}
+		return pipelineRef, pipelineParams, nil
 	}
 
 	// Fallback to the default pipeline
--- a/controllers/component_build_controller_test.go
+++ b/controllers/component_build_controller_test.go
@@ -1802,6 +1802,9 @@ var _ = Describe("Component initial build controller", func() {
 	Context("Test initial build", func() {
 
 		_ = BeforeEach(func() {
+			// For creating build pipeline selector
+			createNamespace(buildServiceNamespaceName)
+
 			ResetTestGitProviderClient()
 
 			pacSecretData := map[string]string{
@@ -1816,6 +1819,7 @@ var _ = Describe("Component initial build controller", func() {
 		_ = AfterEach(func() {
 			deleteSecret(pacSecretKey)
 
+			deleteBuildPipelineRunSelector(defaultSelectorKey)
 			deleteComponentPipelineRuns(resourceKey)
 			deleteComponent(resourceKey)
 		})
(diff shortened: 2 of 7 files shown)


controllers/component_build_controller_common.go
```diff
@@ -68,9 +68,10 @@ func (r *ComponentBuildReconciler) GetPipelineForComponent(ctx context.Context,
...
 		}
-		if pipelineRef != nil {
-			return pipelineRef, pipelineParams, nil
+		if pipelineRef == nil {
+			return nil, nil, boerrors.NewBuildOpError(boerrors.ENoPipelineIsSelected, nil)
 		}
+		return pipelineRef, pipelineParams, nil
 	}
 
 	// Fallback to the default pipeline
```
Michkov, line 77: this fallback should be removed, maybe replaced by error message that no pipeline selector defined.

tkdchen: Updated.

pkg/boerrors/perror.go
```diff
@@ -127,6 +127,9 @@ const (
 	EComponentGitSecretMissing BOErrorId = 201
 	// The secret with image registry credentials specified in 'image.redhat.com/image' annotation does not exist in the user's namespace.
 	EComponentImageRegistrySecretMissing BOErrorId = 202
+
+	ENoPipelineIsSelected            BOErrorId = 300
```
mmorhun, line 131: Please add a description to the errors and how they should be fixed.

controllers/suite_util_test.go
```diff
@@ -547,8 +551,11 @@ func createBuildPipelineRunSelector(selectorKey types.NamespacedName) {
 		Spec: buildappstudiov1alpha1.BuildPipelineSelectorSpec{
 			Selectors: []buildappstudiov1alpha1.PipelineSelector{
 				{
-					Name:           SelectorDefaultName,
-					PipelineRef:    tektonapi.PipelineRef{},
+					Name: SelectorDefaultName,
+					PipelineRef: tektonapi.PipelineRef{
+						Name:   defaultPipelineName,
```
mmorhun, line 556: This util file is using constants from a specific controller test.

controllers/git_tekton_resources_renovater_test.go
```diff
@@ -63,7 +62,7 @@ var _ = Describe("Git tekton resources renovater", func() {
 				return []github.ApplicationInstallation{generateInstallation(repositories)}, "slug", nil
 			}
 			componentNamespacedName := createComponentForPaCBuild(getComponentData(componentConfig{gitURL: "https://github/test/repo3"}))
-			createBuildPipelineRunSelector(defaultSelector)
+			createBuildPipelineRunSelector(defaultSelectorKey)
```
mmorhun, line 65: Shouldn't the selector be created in before each?

controllers/component_build_controller_test.go
```diff
@@ -2016,6 +2029,88 @@ var _ = Describe("Component initial build controller", func() {
...
+
+			component = getComponent(resourceKey)
+			Expect(component.Annotations[InitialBuildAnnotationName]).To(Equal("processed"))
+
+			// Check expected errors
+			statusAnnotation := component.Annotations[BuildStatusAnnotationName]
+			Expect(strings.Contains(statusAnnotation, "No pipeline is selected")).To(BeTrue())
+		})
+
+		// Always keep this test as the last one in current context
```
mmorhun, line 2086: Why do we need this condition? Could we just delete all selectors at the beginning of the test, so there would be no binding any more?

controllers/component_build_controller_test.go
```diff
@@ -1816,6 +1828,7 @@ var _ = Describe("Component initial build controller", func() {
 		_ = AfterEach(func() {
 			deleteSecret(pacSecretKey)
 
+			deleteBuildPipelineRunSelector(defaultSelectorKey)
```
mmorhun, line 1831: Why do we need to delete pipeline selector after each of the initial build test?

controllers/component_build_controller_test.go
```diff
@@ -2178,6 +2273,7 @@ var _ = Describe("Component initial build controller", func() {
 		})
 
 		It("should use the global build bundle", func() {
+			deleteBuildPipelineRunSelector(defaultSelectorKey)
```
mmorhun, line 2276: We do not delete pipeline run selector for the test above, but delete it for this test. Why?
It would be better to avoid fixing order of tests here.

controllers/component_build_controller_test.go
```diff
@@ -2016,6 +2029,88 @@ var _ = Describe("Component initial build controller", func() {
 			ensureNoPipelineRunsCreated(resourceKey)
 		})
 
+		It("should fail when no pipeline is selected based on predefined build pipeline selector", func() {
```
mmorhun, line 2032: Does it make sense to have a similar test for PaC?
Or maybe we should handle error cases for pipeline selection in the pipeline selector test for both? WDYT?

controllers/component_build_controller_test.go
```diff
@@ -2347,4 +2261,99 @@ var _ = Describe("Component initial build controller", func() {
 		})
 	})
 
+	Context("Test pipeline selection", func() {
```
mmorhun, line 2264: This and previous test context look almost identical to me

tkdchen: Updated.

mmorhun (approved):

PR #180, opened 2023-07-18
Update modules and images

Michkov:
- Update images to RHEL9 and golang 1.19
- Update application-service and application-api modules


pkg/pipeline-selector/pipeline_selector_test.go
```diff
@@ -450,7 +450,32 @@ func TestGetPipelineSelectionParametersForComponent(t *testing.T) {
                 components:
                   - name: outerloop-deploy
                     kubernetes:
-                        uri: outerloop-deploy.yaml
```
mkosiarc, line 453: What is the reason behind this change?

Michkov: the devfile validation is more strict now and trying to resolve the file which does not exist.

mkosiarc (approved):

tkdchen (approved):

PR #181, opened 2023-07-18
feat: add configuration time to PaC build status

tnevrlka:
[STONEBLD-1522](https://issues.redhat.com/browse/STONEBLD-1522)
Adds `ConfigurationTime` attribute to `BuildStatus.PaC` which stores the time of last successful PaC configuration in RFC1123 format
Updated relevant tests

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -433,6 +433,7 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 		// Update build status annotation
 		buildStatus := readBuildStatus((&component))
+		pacBuildStatus.ConfigurationTime = buildStatus.PaC.ConfigurationTime
 		buildStatus.PaC = pacBuildStatus
 		buildStatus.Message = "done"
 		writeBuildStatus(&component, buildStatus)
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -145,6 +145,8 @@ type PaCBuildStatus struct {
 	State string `json:"state,omitempty"`
 	// Contains link to PaC provision / unprovision pull request
 	MergeUrl string `json:"merge-url,omitempty"`
+	// Time of the last PaC configuration
+	ConfigurationTime string `json:"configuration-time,omitempty"`
 
 	ErrorInfo
 }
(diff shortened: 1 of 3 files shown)


controllers/component_build_controller.go
```diff
@@ -485,6 +491,9 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 		// Update build status annotation
 		buildStatus := readBuildStatus((&component))
+		if pacBuildStatus.ConfigurationTime == "" && buildStatus.PaC != nil && buildStatus.PaC.ConfigurationTime != "" {
```
Michkov, line 494: not sure if this is needed for unconfigure. That means that configuration-time will be time of when it was unconfigured?

tnevrlka: The way I understand it:

Successfully configure PaC -> `BuildStatus` gets `ConfigurationTime`

Unconfigure PaC -> overwrite whole `buildStatus.PaC` by newly created `pacBuildStatus` which does not have set `ConfigurationTime` attribute, so the previously set `ConfigurationTime` value would be lost and would be empty, which is not intended behavior.

Here, I get the previously set `ConfigurationTime` and give it to the `pacBuildStatus` so it does not get lost

I assume that attribute `ConfigurationTime` becomes empty when the `buildStatus.PaC` is assigned to `pacBuildStatus` on line 497 which does not have that attribute, like in [this example](https://go.dev/play/p/n6cuQ3IYeh8)

Michkov: So that means that in next configuration the time will not be updated because it's already set. Our goal is to have configuration time updated when new configuration is created.
For me it's ok to keep ConfigurationTime not after unconfiguraton.

tnevrlka: There is a condition in the if statement checking that configuration time did not get changed during latest run `pacBuildStatus.ConfigurationTime == ""`
That means it should change the configuration time only if it was not already changed [here on line 415](https://github.com/redhat-appstudio/build-service/pull/181/files/a9800fb2e718b88880e3954b2ae68137082cda32#diff-55dabc319dc9f0a58a4c79d69d79368078f475199412ea72d3fae8bd7c659785R415), shouldn't it?

I see that this code seems confusing, I will try to think of some way to improve it

Michkov (approved):

PR #184, opened 2023-08-01
Provide git repository secret to build pipeline run

mmorhun:
This is a draft implementation that theoretically should work. However, to test it, fixes in Application Service and UI need to be done.


controllers/component_build_controller.go
```diff
@@ -187,6 +187,7 @@ func (r *ComponentBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/status,verbs=get;list;watch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=buildpipelineselectors,verbs=get;list;watch;create;update;patch
+//+kubebuilder:rbac:groups=appstudio.redhat.com,resources=spiaccesstokenbindings,verbs=get;list;watch;update;patch
```
Michkov, line 190: why update/patch?

controllers/component_build_controller_simple_build.go
```diff
@@ -102,6 +67,18 @@ func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component
...
 	}
 
+	// Try to find git secret in case private git repository is used
+	gitSecretName, err := r.findGitSecretName(ctx, component)
+	if err != nil {
+		log.Error(err, "failed to find git secret name")
+		return err
+	}
+	if pRunGitInfo == nil {
+		pRunGitInfo = &pipelineRunGitInfo{}
```
Michkov, line 77: when gitInfo is nil and gitSecretName is empty then it should fail with error on component. So that HAC could trigger OAuth workflow and retry.

Michkov (changes requested):

controllers/component_build_controller.go
mmorhun: Not needed according to the latest solution, removing.

rcerven:
and what about custom wf in https://github.com/redhat-appstudio/build-service/blob/main/controllers/component_build_controller_pac.go#L761 ?

rcerven:
> @rcerven I didn't get your point, could you elaborate?

well don't you have to set that secret also for custom workflow in that place I've pasted?

mmorhun:
@rcerven no, Pipelines-as-Code will take care about credentials. We need the secret for simple builds only.

controllers/component_build_controller_simple_build.go
```diff
@@ -168,19 +159,43 @@ func getPipelineRunGitInfo(component *appstudiov1alpha1.Component, pacConfig map
 	if gitSourceSha == "" {
 		gitSourceSha, err = gitClient.GetBranchSha(repoUrl, revision)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get git branch SHA: %w", err)
+			log.Error(err, "failed to get git branch SHA")
```
tkdchen, line 162: Why the error can be ignored?

If ignore this error, `browseRepositoryAtShaLink` will be empty eventually, how should it be treated, error or ignored as well?

If this error can be ignored, it would be good to mention this info in the message.

controllers/component_build_controller_simple_build.go
```diff
@@ -128,13 +90,18 @@ func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component
 	return nil
 }
 
-type pipelineRunGitInfo struct {
+type buildGitInfo struct {
+	isPublic      bool
+	gitSecretName string
```
tkdchen, line 95: Could you add a comment for this field?

controllers/component_build_controller_simple_build.go
mmorhun: > Why the error can be ignored?

There is comment above. In short they are needed only for UI.

> If ignore this error, browseRepositoryAtShaLink will be empty eventually

That's expected, again this needed only for UI.

controllers/component_build_controller_simple_build.go
mmorhun: ok

Michkov (approved):

tkdchen (approved):

PR #192, opened 2023-08-25
Increase test coverage in git_tekton_resources_renovater.go

rcerven:
When checking PACsecret existence, don't filter out NotFound error, as it would fail eventually anyway.

Test for non-existing PACsecret has to be the last, so other test don't conflict with it and won't create secret for it.

[STONEBLD-1590](https://issues.redhat.com//browse/STONEBLD-1590)

### DIFF
--- a/controllers/git_tekton_resources_renovater.go
+++ b/controllers/git_tekton_resources_renovater.go
@@ -114,11 +114,13 @@ func (r *GitTektonResourcesRenovater) Reconcile(ctx context.Context, req ctrl.Re
 	pacSecret := corev1.Secret{}
 	globalPaCSecretKey := types.NamespacedName{Namespace: buildServiceNamespaceName, Name: gitopsprepare.PipelinesAsCodeSecretName}
 	if err := r.Client.Get(ctx, globalPaCSecretKey, &pacSecret); err != nil {
-		if !errors.IsNotFound(err) {
-			r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
+		r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
+		if errors.IsNotFound(err) {
+			log.Error(err, "not found Pipelines as Code secret in %s namespace: %w", globalPaCSecretKey.Namespace, err, l.Action, l.ActionView)
+		} else {
 			log.Error(err, "failed to get Pipelines as Code secret in %s namespace: %w", globalPaCSecretKey.Namespace, err, l.Action, l.ActionView)
-			return ctrl.Result{}, nil
 		}
+		return ctrl.Result{}, nil
 	}
 	isApp := gitops.IsPaCApplicationConfigured("github", pacSecret.Data)
 	if !isApp {
--- a/controllers/git_tekton_resources_renovater_test.go
+++ b/controllers/git_tekton_resources_renovater_test.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"os"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -50,6 +49,7 @@ var _ = Describe("Git tekton resources renovater", func() {
 			deleteBuildPipelineRunSelector(defaultSelectorKey)
 			deleteJobs(buildServiceNamespaceName)
 			os.Unsetenv(InstallationsPerJobEnvName)
+			deleteSecret(pacSecretKey)
 		})
 
 		It("It should not trigger job", func() {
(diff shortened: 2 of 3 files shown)


tkdchen (approved):

controllers/git_tekton_resources_renovater_test.go
```diff
@@ -61,11 +62,12 @@ var _ = Describe("Git tekton resources renovater", func() {
 				repositories := generateRepositories(installedRepositoryUrls)
 				return []github.ApplicationInstallation{generateInstallation(repositories)}, "slug", nil
 			}
-			componentNamespacedName := createComponentForPaCBuild(getComponentData(componentConfig{gitURL: "https://github/test/repo3"}))
+			componentNamespacedName := createComponentForPaCBuild(getComponentData(componentConfig{componentKey: types.NamespacedName{Name: "testnotrigger"}, gitURL: "https://github/test/repo3"}))
```
mmorhun, line 65: What is the reason to change the component name?

controllers/git_tekton_resources_renovater.go
```diff
@@ -114,11 +113,9 @@ func (r *GitTektonResourcesRenovater) Reconcile(ctx context.Context, req ctrl.Re
 	pacSecret := corev1.Secret{}
 	globalPaCSecretKey := types.NamespacedName{Namespace: buildServiceNamespaceName, Name: gitopsprepare.PipelinesAsCodeSecretName}
 	if err := r.Client.Get(ctx, globalPaCSecretKey, &pacSecret); err != nil {
-		if !errors.IsNotFound(err) {
-			r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
-			log.Error(err, "failed to get Pipelines as Code secret in %s namespace: %w", globalPaCSecretKey.Namespace, err, l.Action, l.ActionView)
-			return ctrl.Result{}, nil
-		}
+		r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
```
mmorhun, line 116: It would be better to distinguish between secret does not exist (in fact not configured) and any other errors. At least with the error message.

controllers/git_tekton_resources_renovater_test.go
```diff
@@ -75,11 +77,30 @@ var _ = Describe("Git tekton resources renovater", func() {
...
+			github.GetAppInstallations = func(appIdStr string, privateKeyPem []byte) ([]github.ApplicationInstallation, string, error) {
+				repositories := generateRepositories(installedRepositoryUrls)
+				return []github.ApplicationInstallation{generateInstallation(repositories)}, "slug", nil
+			}
+			componentsData := getComponentData(componentConfig{componentKey: types.NamespacedName{Name: "testdefault"}, gitURL: "https://github/test/repo1"})
+			componentsData.Spec.Source.GitSource.Revision = ""
+			componentNamespacedName := createComponentForPaCBuild(componentsData)
+			os.Setenv(InstallationsPerJobEnvName, "0")
+			createBuildPipelineRunSelector(defaultSelectorKey)
+			Eventually(listJobs).WithArguments(buildServiceNamespaceName).WithTimeout(10 * time.Second).Should(HaveLen(1))
```
mmorhun, line 100: Why not to use globally defined `timeout` constant?

controllers/git_tekton_resources_renovater_test.go
```diff
@@ -127,5 +148,24 @@ var _ = Describe("Git tekton resources renovater", func() {
 			deleteComponent(componentNamespacedName3)
 			deleteComponent(componentNamespacedName4)
 		})
+
+		// test has to be the last, because it isn't waiting for reconcile action,
+		// so test which would ran after it would create secret again
+		It("It should trigger job, because pac secret is missing", func() {
```
mmorhun, line 154: What does this test check? I see no asserts in it.

controllers/git_tekton_resources_renovater_test.go
rcerven: changed them back

controllers/git_tekton_resources_renovater.go
rcerven: logging different message for both cases

controllers/git_tekton_resources_renovater_test.go
rcerven: 10 seconds was there for all other tests,

changed it everywhere to 'timeout'

controllers/git_tekton_resources_renovater_test.go
rcerven: asserting event now

controllers/git_tekton_resources_renovater.go
```diff
@@ -114,11 +114,13 @@ func (r *GitTektonResourcesRenovater) Reconcile(ctx context.Context, req ctrl.Re
 	pacSecret := corev1.Secret{}
 	globalPaCSecretKey := types.NamespacedName{Namespace: buildServiceNamespaceName, Name: gitopsprepare.PipelinesAsCodeSecretName}
 	if err := r.Client.Get(ctx, globalPaCSecretKey, &pacSecret); err != nil {
+		r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
 		if !errors.IsNotFound(err) {
-			r.EventRecorder.Event(&pacSecret, "Warning", "ErrorReadingPaCSecret", err.Error())
 			log.Error(err, "failed to get Pipelines as Code secret in %s namespace: %w", globalPaCSecretKey.Namespace, err, l.Action, l.ActionView)
-			return ctrl.Result{}, nil
+		} else {
```
mmorhun, line 120: I'd avoid negation in the `if`since `else` is here as well.

controllers/git_tekton_resources_renovater_test.go
```diff
@@ -121,11 +141,32 @@ var _ = Describe("Git tekton resources renovater", func() {
...
 			createBuildPipelineRunSelector(defaultSelectorKey)
-			Eventually(listJobs).WithArguments(buildServiceNamespaceName).WithTimeout(10 * time.Second).Should(HaveLen(2))
+			Eventually(listJobs).WithArguments(buildServiceNamespaceName).WithTimeout(timeout).Should(HaveLen(2))
 			deleteComponent(componentNamespacedName1)
 			deleteComponent(componentNamespacedName2)
 			deleteComponent(componentNamespacedName3)
 			deleteComponent(componentNamespacedName4)
 		})
+
+		// test has to be the last, because it isn't waiting for reconcile action,
```
mmorhun, line 151: After using assertion, I suppose, this test doesn't have to be the latest.

controllers/git_tekton_resources_renovater_test.go
mmorhun: Which exactly? Can we do proper cleanup?

PR #194, opened 2023-08-30
Reuse existing PaC repository CR for multi component git repositories

mmorhun:
This PR changes the way how PaC repository CR is handled. Instead of creation of a new PaC repository per component, Build Service checks if the is existing matching PaC repository and if so, uses it.

Also, error handling logic is added to stop if PaC admission webhook denied PaC repository object creation:
```yaml
build.appstudio.openshift.io/status: '{"pac":{"state":"error","error-id":53,"error-message":"53: Git repository is already handled by Pipelines as Code"},"message":"done"}'
```

Tested using [this](https://github.com/mmorhun/sample-multi-component) sample multi component git repository.

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -182,7 +182,7 @@ func (r *ComponentBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=buildpipelineselectors,verbs=get;list;watch;create;update;patch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=spiaccesstokenbindings,verbs=get;list;watch
 //+kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns,verbs=create
-//+kubebuilder:rbac:groups=pipelinesascode.tekton.dev,resources=repositories,verbs=get;list;watch;create;update;patch
+//+kubebuilder:rbac:groups=pipelinesascode.tekton.dev,resources=repositories,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch;create;patch;update
 //+kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch
--- a/controllers/component_build_controller_pac.go
+++ b/controllers/component_build_controller_pac.go
@@ -44,6 +44,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/yaml"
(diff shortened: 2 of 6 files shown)


mmorhun:
Unfortunately, it's not possible to cover PaC repository error handling with integration tests. Real cluster is required to do so.

controllers/component_build_controller_pac.go
```diff
@@ -407,6 +408,29 @@ func checkMandatoryFieldsNotEmpty(config map[string][]byte, mandatoryFields []st
...
+				log.Error(err, "failed to add owner reference to existing PaC repository", "PaCRepositoryName", repository.Name)
+				return err
+			}
+			if len(repository.OwnerReferences) > pacRepositoryOwnersNumber {
+				if err := r.Client.Update(ctx, repository); err != nil {
+					log.Error(err, "failed to update existing PaC repository with component owner reference", "PaCRepositoryName", repository.Name)
+					return err
+				}
+				log.Info("Added current component to owners of the PaC repository", "PaCRepositoryName", repository.Name, l.Action, l.ActionUpdate)
+			}
```
rcerven, line 429: else, should return error

and remove that return nil

controllers/component_build_controller_pac.go
```diff
@@ -419,6 +443,12 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
 				return err
 			}
 			if err := r.Client.Create(ctx, repository); err != nil {
+				if strings.Contains(err.Error(), "repository already exist with url") {
```
rcerven, line 446: k8sErrors.IsAlreadyExists  should work?

controllers/component_build_controller_pac.go
mmorhun: I believe, that the current logic is correct. We should not return error is everything is already up to date.

controllers/component_build_controller_pac.go
mmorhun: I think `IsAlreadyExists` will not work in this case.

controllers/component_build_controller_pac.go
tkdchen, line 429: My understanding is that, `if len(repository.OwnerReferences) > pacRepositoryOwnersNumber {` is for checking if the owner references is updated with another component within subdir in the same git repository. If yes, update the PaC Repository object.

[SetOwnerReference](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetOwnerReference) document mentions this

> If a reference to the same object already exists, it'll be overwritten with the newly provided version. 

my understanding is, in whatever the case, the found PaC Repository is modified by calling `SetOwnerReference`. If we don't call `Client.Update`, will there be an issue that the Repository object is modified but not updated. We can probably always call `Client.Update` to avoid any potential issues.

controllers/component_build_controller_pac.go
tkdchen, line 446: Above code has been handled the multi-components case, and here the component should not have a PaC Repository yet. Why check this error again here, would it happen again in some cases?

controllers/component_build_controller_pac.go
mmorhun: Yes, you are completely correct.

> If we don't call Client.Update, will there be an issue that the Repository object is modified but not updated. 

I don't think so, because such situation is not possible.
We have 4 fields in the owner reference:
 - `APIVersion` which is a constant for our use case.
 - `Kind` is a constant as well.
 - `Name` of the Component cannot be changed on the go.
 - `UID` is the constant by definition.
So, such a situation is not possible.

tkdchen (approved):

mkosiarc (approved):

controllers/component_build_controller_pac.go
rcerven: I see, ok

controllers/component_build_controller_pac.go
rcerven: I see, because that means that reference was already there

rcerven (approved):

controllers/component_build_controller.go
```diff
@@ -182,7 +182,7 @@ func (r *ComponentBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=buildpipelineselectors,verbs=get;list;watch;create;update;patch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=spiaccesstokenbindings,verbs=get;list;watch
 //+kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns,verbs=create
-//+kubebuilder:rbac:groups=pipelinesascode.tekton.dev,resources=repositories,verbs=get;list;watch;create;update;patch
+//+kubebuilder:rbac:groups=pipelinesascode.tekton.dev,resources=repositories,verbs=get;list;watch;create;update;patch;delete
```
Michkov, line 185: why delete has been added? Deletion is still managed by ownershipReference.

mmorhun: Because we cannot set owner reference unless we have deletion rights.

PR #196, opened 2023-09-08
Increase test coverage in component_build_controller_pac.go and compo…

rcerven:
…nent_build_controller_simple_build.go

Properly handle errors when:
gitprovider can't be detected
pac secret doesn't exist

When UndoPaCProvisionForComponent fails with persistent error set buildStatus.State to error

Added 1 sec sleep in AfterEach after component deletion, so it waits for pruner

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -50,8 +50,6 @@ const (
 
 	BuildStatusAnnotationName = "build.appstudio.openshift.io/status"
 
-	InitialBuildAnnotationName = "appstudio.openshift.io/component-initial-build"
-
 	PaCProvisionFinalizer            = "pac.component.appstudio.openshift.io/finalizer"
 	ImageRegistrySecretLinkFinalizer = "image-registry-secret-sa-link.component.appstudio.openshift.io/finalizer"
 
@@ -213,9 +211,9 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
-	// Do not run any builds for any container-image components
-	if component.Spec.ContainerImage != "" && (component.Spec.Source.GitSource == nil || component.Spec.Source.GitSource.URL == "") {
-		log.Info("Nothing to do for container image component")
+	// Do not run any builds if component doesn't have gitsource
+	if component.Spec.Source.GitSource == nil || component.Spec.Source.GitSource.URL == "" {
+		log.Info("Nothing to do for container image component without source")
 		return ctrl.Result{}, nil
 	}
 
@@ -354,8 +352,6 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		buildStatus.Message = "done"
 		writeBuildStatus(&component, buildStatus)
 
-		component.Annotations[InitialBuildAnnotationName] = "processed"
-
 	case BuildRequestConfigurePaCAnnotationValue:
 		pacBuildStatus := &PaCBuildStatus{}
 		if mergeUrl, err := r.ProvisionPaCForComponent(ctx, &component); err != nil {
@@ -417,6 +413,7 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		if mergeUrl, err := r.UndoPaCProvisionForComponent(ctx, &component); err != nil {
 			if boErr, ok := err.(*boerrors.BuildOpError); ok && boErr.IsPersistent() {
 				log.Error(err, "Pipelines as Code unprovision for the Component failed")
+				pacBuildStatus.State = "error"
 				pacBuildStatus.ErrId = boErr.GetErrorId()
 				pacBuildStatus.ErrMessage = boErr.ShortError()
 			} else {
--- a/controllers/component_build_controller_pac.go
+++ b/controllers/component_build_controller_pac.go
@@ -158,14 +158,14 @@ func (r *ComponentBuildReconciler) UndoPaCProvisionForComponent(ctx context.Cont
 	if err != nil {
 		log.Error(err, "error detecting git provider")
 		// There is no point to continue if git provider is not known.
-		return "", err
+		return "", boerrors.NewBuildOpError(boerrors.EUnknownGitProvider, err)
 	}
 
 	pacSecret := corev1.Secret{}
 	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: buildServiceNamespaceName, Name: gitopsprepare.PipelinesAsCodeSecretName}, &pacSecret); err != nil {
 		log.Error(err, "error getting git provider credentials secret", l.Action, l.ActionView)
 		// Cannot continue without accessing git provider credentials.
-		return "", err
+		return "", boerrors.NewBuildOpError(boerrors.EPaCSecretNotFound, err)
 	}
 
 	webhookTargetUrl := ""
--- a/controllers/component_build_controller_simple_build.go
+++ b/controllers/component_build_controller_simple_build.go
@@ -56,7 +56,7 @@ func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component
 	pacSecret := corev1.Secret{}
 	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: buildServiceNamespaceName, Name: gitopsprepare.PipelinesAsCodeSecretName}, &pacSecret); err != nil {
 		log.Error(err, "failed to get git provider credentials secret", l.Action, l.ActionView)
-		return err
+		return boerrors.NewBuildOpError(boerrors.EPaCSecretNotFound, err)
 	}
 	buildGitInfo, err := r.getBuildGitInfo(ctx, component, pacSecret.Data)
 	if err != nil {
@@ -107,7 +107,8 @@ func (r *ComponentBuildReconciler) getBuildGitInfo(ctx context.Context, componen
 	gitProvider, err := gitops.GetGitProvider(*component)
 	if err != nil {
 		// There is no point to continue if git provider is not known
-		return nil, fmt.Errorf("error detecting git provider: %w", err)
+		log.Error(err, "error detecting git provider")
+		return nil, boerrors.NewBuildOpError(boerrors.EUnknownGitProvider, err)
 	}
 
 	repoUrl := component.Spec.Source.GitSource.URL
--- a/controllers/component_build_controller_test.go
+++ b/controllers/component_build_controller_test.go
@@ -32,9 +32,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	"github.com/devfile/api/v2/pkg/apis/workspaces/v1alpha2"
 	appstudiov1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/application-service/gitops"
 	gitopsprepare "github.com/redhat-appstudio/application-service/gitops/prepare"
+	"github.com/redhat-appstudio/application-service/pkg/devfile"
 	buildappstudiov1alpha1 "github.com/redhat-appstudio/build-service/api/v1alpha1"
 	"github.com/redhat-appstudio/build-service/pkg/boerrors"
 	"github.com/redhat-appstudio/build-service/pkg/git/github"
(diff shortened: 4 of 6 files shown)


tkdchen:
> Added 1 sec sleep in AfterEach after component deletion, so it waits for pruner

What would happen if not wait for the pruner?

controllers/component_build_controller_simple_build.go
```diff
@@ -66,7 +66,7 @@ func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component
 	buildPipelineRun, err := generatePipelineRunForComponent(component, pipelineRef, additionalPipelineParams, buildGitInfo)
 	if err != nil {
 		log.Error(err, fmt.Sprintf("Failed to generate PipelineRun to build %s component in %s namespace", component.Name, component.Namespace))
-		return err
+		return boerrors.NewBuildOpError(boerrors.EUnknownError, err)
```
tkdchen, line 69: This change is unnecessary. The error returned from `generatePipelineRunForComponent` is already a persistent error `EInvalidDevfile`.

rcerven:
> > Added 1 sec sleep in AfterEach after component deletion, so it waits for pruner
> 
> What would happen if not wait for the pruner?

tests are randomly failing due that race

controllers/component_build_controller_simple_build.go
rcerven: ahh right that was recently added by Mykola, will remove this

controllers/component_build_controller_test.go
```diff
@@ -761,6 +763,104 @@ var _ = Describe("Component initial build controller", func() {
 			deleteRoute(pacRouteKey)
 		})
 
+		It("should not submit initial build if pac secret is missing", func() {
```
mmorhun, line 766: Looks like the test name is wrong

controllers/component_build_controller_test.go
```diff
@@ -1148,7 +1249,27 @@ var _ = Describe("Component initial build controller", func() {
...
+			ensureComponentInitialBuildAnnotationState(resourceKey, true)
+
+			// Check pipeline run labels and annotations
+			pipelineRun := listComponentPipelineRuns(resourceKey)[0]
+			Expect(pipelineRun.Annotations[gitCommitShaAnnotationName]).To(Equal(gitSourceSHA))
+			Expect(pipelineRun.Annotations[gitRepoAtShaAnnotationName]).To(Equal(DefaultBrowseRepository + gitSourceSHA))
+		})
+
 		It("should be able to retrigger simple build", func() {
+
```
mmorhun, line 1272: nitpick: why newline here?

controllers/component_build_controller_test.go
```diff
@@ -1288,6 +1409,12 @@ var _ = Describe("Component initial build controller", func() {
 			ensureNoPipelineRunsCreated(resourceKey)
 		})
 
+		It("should not submit initial build if pac secret is missing", func() {
+			deleteSecret(pacSecretKey)
+			setComponentDevfileModel(resourceKey)
+			ensureNoPipelineRunsCreated(resourceKey)
```
mmorhun, line 1415: I'd rather check status annotation instead of using `ensureNoPipelineRunsCreated`, where possible.

controllers/component_build_controller_test.go
```diff
@@ -1330,6 +1457,52 @@ var _ = Describe("Component initial build controller", func() {
...
 			ensureNoPipelineRunsCreated(resourceKey)
 		})
+
+		It("should not submit initial build if Dockerfile is wrong", func() {
+			DevfileSearchForDockerfile = func(devfileBytes []byte) (*v1alpha2.DockerfileImage, error) {
+				return nil, fmt.Errorf("wrong dockerfile")
+			}
+			setComponentDevfileModel(resourceKey)
+			ensureNoPipelineRunsCreated(resourceKey)
+			DevfileSearchForDockerfile = devfile.SearchForDockerfile
```
mmorhun, line 1467: This is fragile because the test might fail before reaching this line.

controllers/component_build_controller_test.go
```diff
@@ -1349,6 +1522,8 @@ var _ = Describe("Component initial build controller", func() {
 			deleteComponent(resourceKey)
 			deleteComponentPipelineRuns(resourceKey)
 			deleteSecret(pacSecretKey)
+			// wait for pruner operator to finish, so it won't prune runs from new test
+			time.Sleep(time.Second)
```
mmorhun, line 1526: Have you got an idea why it works without sleep for the whole file?

controllers/component_build_controller.go
```diff
@@ -213,9 +213,9 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
-	// Do not run any builds for any container-image components
-	if component.Spec.ContainerImage != "" && (component.Spec.Source.GitSource == nil || component.Spec.Source.GitSource.URL == "") {
-		log.Info("Nothing to do for container image component")
+	// Do not run any builds if gitsource is missing from component
```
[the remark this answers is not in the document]
rcerven: ok will rephrase

controllers/component_build_controller_pac.go
```diff
@@ -158,14 +158,15 @@ func (r *ComponentBuildReconciler) UndoPaCProvisionForComponent(ctx context.Cont
...
 		log.Error(err, "error detecting git provider")
 		// There is no point to continue if git provider is not known.
-		return "", err
+		return "", boerrors.NewBuildOpError(boerrors.EUnknownGitProvider, err)
 	}
 
 	pacSecret := corev1.Secret{}
 	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: buildServiceNamespaceName, Name: gitopsprepare.PipelinesAsCodeSecretName}, &pacSecret); err != nil {
 		log.Error(err, "error getting git provider credentials secret", l.Action, l.ActionView)
+
```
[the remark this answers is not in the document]
rcerven: not really removing

controllers/component_build_controller_test.go
rcerven: why? I am removing pac secret and build will fail

controllers/component_build_controller_test.go
rcerven: removing

controllers/component_build_controller_test.go
rcerven: ok will add

controllers/component_build_controller_test.go
rcerven: ok moving to beforeEach

controllers/component_build_controller_test.go
mmorhun: First, you're testing PaC not initial build. Second, it doesn't build, it's doing PaC unprovision.

controllers/component_build_controller_test.go
```diff
@@ -1288,6 +1413,20 @@ var _ = Describe("Component initial build controller", func() {
 			ensureNoPipelineRunsCreated(resourceKey)
 		})
 
+		It("should not submit initial build if pac secret is missing", func() {
+			deleteSecret(pacSecretKey)
+			setComponentDevfileModel(resourceKey)
+			ensureNoPipelineRunsCreated(resourceKey)
```
mmorhun, line 1419: My concern with `ensure*` methods is that they use `Consistently` under the hood (unlike `wait*` that use `Eventually`), which means each invocation takes `ensureTimeout` i.e. `4` seconds.
I don't mind they're used, but would like to do it reasonable to keep total testing time minimal (but not sacrificing quality, of course).

controllers/component_build_controller_test.go
tkdchen: Could one component per test be helpful? Or, launch the pruner controller only before tests that test pruner?

controllers/component_build_controller_test.go
```diff
@@ -1330,6 +1469,65 @@ var _ = Describe("Component initial build controller", func() {
...
+		})
+
+		It("should not submit initial build if ContainerImage is not set)", func() {
+			deleteComponent(resourceKey)
+
+			component := getSampleComponentData(resourceKey)
+			component.Spec.ContainerImage = ""
+
+			Expect(k8sClient.Create(ctx, component)).Should(Succeed())
```
tkdchen, line 1526: Would be good to wait for the component creation to finish explicitly.

controllers/component_build_controller_test.go
rcerven: it would help for sure, but I am not going to change now everything

controllers/component_build_controller_test.go
rcerven: ok will rephrase :)

rcerven:
> > Added 1 sec sleep in AfterEach after component deletion, so it waits for pruner
> 
> What would happen if not wait for the pruner?

well problem is that we have pruner controller active, so when you delete component, it triggers ALSO pruner, at that time already runs new test which creates component

and now there can be race between pruner cleaning OLD/NEW pipeline runs

controllers/component_build_controller_test.go
rcerven: ok

controllers/component_build_controller_test.go
```diff
@@ -1148,6 +1254,25 @@ var _ = Describe("Component initial build controller", func() {
...
+
+			gitSourceSHA := "d1a9e858489d1515621398fb02942da068f1c956"
+			component := getSampleComponentData(resourceKey)
+			component.Spec.Source.GitSource.Revision = gitSourceSHA
+			Expect(k8sClient.Create(ctx, component)).Should(Succeed())
+			setComponentDevfileModel(resourceKey)
+
+			waitOneInitialPipelineRunCreated(resourceKey)
+			waitComponentAnnotationGone(resourceKey, BuildRequestAnnotationName)
+			ensureComponentInitialBuildAnnotationState(resourceKey, true)
```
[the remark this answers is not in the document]
rcerven: ok

controllers/component_build_controller_test.go
rcerven: ok I'll check where I can avoid it

controllers/component_build_controller_test.go
```diff
@@ -1288,6 +1413,20 @@ var _ = Describe("Component initial build controller", func() {
 			ensureNoPipelineRunsCreated(resourceKey)
 		})
 
+		It("should not submit initial build if pac secret is missing", func() {
+			deleteSecret(pacSecretKey)
```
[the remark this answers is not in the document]
rcerven: yes, but I am testing missing secret, and since BeforeEach creates secret, I have to remove it

controllers/component_build_controller_test.go
rcerven: yeah added it everywhere, as it might be causing also race with update

controllers/component_build_controller_test.go
```diff
@@ -1459,6 +1653,7 @@ var _ = Describe("Component initial build controller", func() {
 				},
 			}
 			Expect(k8sClient.Create(ctx, selectors)).To(Succeed())
+			getComponent(resourceKey)
```
mmorhun, line 1656: Why do we need to wait for component after creation of the selectors?

controllers/component_build_controller_test.go
mmorhun: Waiting for component is not added here.

controllers/component_build_controller_test.go
rcerven: mistake, removed :)

rcerven: and added to another

controllers/component_build_controller_test.go
```diff
@@ -578,9 +580,7 @@ var _ = Describe("Component initial build controller", func() {
 		It("should set error in status if invalid build action requested", func() {
 			createCustomComponentWithBuildRequest(componentConfig{
 				componentKey: resourceKey,
-				annotations: map[string]string{
-					InitialBuildAnnotationName: "processed",
-				},
+				annotations:  map[string]string{},
```
mmorhun, line 583: This line can be removed

controllers/component_build_controller_test.go
```diff
@@ -761,6 +761,108 @@ var _ = Describe("Component initial build controller", func() {
 			deleteRoute(pacRouteKey)
 		})
 
+		It("should fail to unconfigure PaC when pac secret is missing", func() {
```
mmorhun, line 764: Usually the most common successful use case is the first test.
Also, successful cases goes first, then failure cases. However there are exceptions if the test cases are grouped logically. Please try to follow this convention.

mmorhun (approved):
Good work!

controllers/component_build_controller_test.go
rcerven: ack reordered

controllers/component_build_controller_test.go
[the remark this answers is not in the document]
tkdchen: You are right. Sorry for my invalid comment.

controllers/component_build_controller_test.go
tkdchen: I think we should consider the refactor on component creation functions in order to avoid doing the create and wait pattern separately. This is out of the scope of this PR.

controllers/component_build_controller_test.go
tkdchen: > it would help for sure, but I am not going to change now everything

Yeah, this can be done separately.

PR #198, opened 2023-09-13
Refactor component_build_controller_test.go

rcerven:
Use different componet name for each context

New helper methods for:
creating custom component from privided data, with wait for component
check pac buildstatus with commonly checked keys
check simple buildstatus with commonly checked keys

### DIFF
--- a/controllers/suite_util_test.go
+++ b/controllers/suite_util_test.go
@@ -214,9 +214,19 @@ func createComponentForPaCBuild(sampleComponentData *appstudiov1alpha1.Component
 func createComponent(componentLookupKey types.NamespacedName) {
 	component := getSampleComponentData(componentLookupKey)
 
-	Expect(k8sClient.Create(ctx, component)).Should(Succeed())
+	createComponentCustom(component)
+}
 
-	getComponent(componentLookupKey)
+// createComponentCustom creates custom component resource and verifies it was properly created
+func createComponentCustom(sampleComponentData *appstudiov1alpha1.Component) {
+	Expect(k8sClient.Create(ctx, sampleComponentData)).Should(Succeed())
+
+	lookupKey := types.NamespacedName{
+		Name:      sampleComponentData.Name,
+		Namespace: sampleComponentData.Namespace,
+	}
+
+	getComponent(lookupKey)
 }
 
 func getComponent(componentKey types.NamespacedName) *appstudiov1alpha1.Component {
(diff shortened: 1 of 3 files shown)


controllers/suite_util_test.go
```diff
@@ -300,6 +312,31 @@ func waitDoneMessageOnComponent(componentKey types.NamespacedName) {
...
+	}
+}
+
+func expectSimpleBuildStatus(componentKey types.NamespacedName, errID int, errMessage string, startTimeEmpty bool) {
+	buildStatus := readBuildStatus(getComponent(componentKey))
+	Expect(buildStatus).ToNot(BeNil())
+	Expect(buildStatus.Simple).ToNot(BeNil())
+	Expect(buildStatus.Simple.ErrId).To(Equal(errID))
+	Expect(buildStatus.Simple.ErrMessage).To(Equal(errMessage))
+	if startTimeEmpty == true {
```
tkdchen, line 333: `if startTimeEmpty` should be enough

controllers/suite_util_test.go
```diff
@@ -300,6 +312,31 @@ func waitDoneMessageOnComponent(componentKey types.NamespacedName) {
...
+func expectPacBuildStatus(componentKey types.NamespacedName, state string, errID int, errMessage string, mergeURL string) {
+	buildStatus := readBuildStatus(getComponent(componentKey))
+	Expect(buildStatus).ToNot(BeNil())
+	Expect(buildStatus.PaC).ToNot(BeNil())
+	Expect(buildStatus.PaC.State).To(Equal(state))
+	Expect(buildStatus.PaC.ErrId).To(Equal(errID))
+	Expect(buildStatus.PaC.ErrMessage).To(Equal(errMessage))
+	if mergeURL != "" {
+		Expect(buildStatus.PaC.MergeUrl).To(Equal(mergeURL))
+	}
```
tkdchen, line 324: Arguments `state`, `errID`, `errMessage` and `mergeURL` reference the expected values. We should always compare them with the corresponding value got from the component.

controllers/suite_util_test.go
```diff
@@ -300,6 +312,31 @@ func waitDoneMessageOnComponent(componentKey types.NamespacedName) {
...
+}
```
tkdchen, line 325: `ConfigurationTime` is not included in this function.

controllers/suite_util_test.go
```diff
@@ -219,6 +219,18 @@ func createComponent(componentLookupKey types.NamespacedName) {
...
+func createComponentCustom(sampleComponentData *appstudiov1alpha1.Component) {
+	Expect(k8sClient.Create(ctx, sampleComponentData)).Should(Succeed())
+
+	lookupKey := types.NamespacedName{
+		Name:      sampleComponentData.Name,
+		Namespace: sampleComponentData.Namespace,
+	}
+
+	getComponent(lookupKey)
+}
```
tkdchen, line 232: Ideally, this `createComponentCustom` and existing `createComponent` are merged eventually.

tkdchen (commented):
Nice. Add a few comments.

controllers/suite_util_test.go
rcerven: I did make createComponent use createComponentCustom

one needs just key another whole component, so merging would mean cover both cases, and I think it is better to have that function simpler

controllers/suite_util_test.go
rcerven: ok I'll make it dependent on state

if enabled configurationtime is there
otherwise it is empty

mmorhun (approved):

PR #199, opened 2023-09-15
Improve coverage in component_build_controller

rcerven:
also fix:
when unlinkSecretFromServiceAccount, when cleaning ImagePullSecrets don't rely on length of Secrets but ImagePullSecrets

[STONEBLD-1590](https://issues.redhat.com//browse/STONEBLD-1590)

### DIFF
--- a/controllers/suite_util_test.go
+++ b/controllers/suite_util_test.go
@@ -311,6 +311,19 @@ func waitDoneMessageOnComponent(componentKey types.NamespacedName) {
 }
 
 func expectPacBuildStatus(componentKey types.NamespacedName, state string, errID int, errMessage string, mergeURL string) {
+	// in 1 test component is usually created (which triggers reconcile and adds message=done)
+	// and then component is updated (and waits for message=done apart from other things)
+	// we should wait for desired state as well, as there is small time window when
+	// message might be still done from previous reconcile, but even though requested action already finished,
+	// it didn't update status yet so it will get status from previous reconcile
+	Eventually(func() bool {
+		buildStatus := readBuildStatus(getComponent(componentKey))
+		Expect(buildStatus).ToNot(BeNil())
+		Expect(buildStatus.PaC).ToNot(BeNil())
+
+		return buildStatus.PaC.State == state
+	}, timeout, interval).Should(BeTrue())
+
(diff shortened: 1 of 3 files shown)


controllers/suite_util_test.go
```diff
@@ -311,6 +311,17 @@ func waitDoneMessageOnComponent(componentKey types.NamespacedName) {
 }
 
 func expectPacBuildStatus(componentKey types.NamespacedName, state string, errID int, errMessage string, mergeURL string) {
+	// should first wait for desired state, as there is small time window when
+	// message might be still done from previous run, but even though action already finished, it didn't update status yet
+	// so it will get status from previous run
```
tkdchen, line 316: Regarding the "previous run", do you mean a previous test in same context might be still running while current test is running?

rcerven: no, but previous test ran and finished and set
message=done
and pac state=enabled

problem is that when testing failed build
we can wait for finilazer to be gone
message = done
but there is still time between that and actually updating component with new build status

so waiting at least on desired state will help with the race

tkdchen (approved):

mkosiarc:
I would prefer if the fix and the coverage increase were in separate commits.

controllers/suite_util_test.go
```diff
@@ -311,6 +311,17 @@ func waitDoneMessageOnComponent(componentKey types.NamespacedName) {
 }
 
 func expectPacBuildStatus(componentKey types.NamespacedName, state string, errID int, errMessage string, mergeURL string) {
+	// should first wait for desired state, as there is small time window when
+	// message might be still done from previous run, but even though action already finished, it didn't update status yet
```
mmorhun, line 315: > message might be still done from previous run

I believe this should never happen as we recreate the Component in each test case.
If you mean checking Component state within the same test case, please make the message clearer (at least replace run with check).

rcerven:
> I would prefer if the fix and the coverage increase were in separate commits.

done

controllers/suite_util_test.go
rcerven: rephrased

mmorhun (approved):

PR #204, opened 2023-10-16
Use cel expression for build pipeline event filtering

mmorhun:
This PR prevents redundant builds for multi component git repositories. A new build is started only if files related to the Component are changed.


tkdchen (commented):
LGTM. A minor suggestion is to use `strings.Join` to join the expression conditions.

rcerven:
needs also added ".tekton" (with specific pruns) in the filter, so change only to pruns will actually trigger builds as well

rcerven (approved):

PR #205, opened 2023-10-27
Migrate to Tekton API v1 while maintaining backwards compatibility

chmeliik:
Currently, build-service only supports v1beta1 style build selectors that resolve to v1beta1 pipelines

```yaml
kind: BuildPipelineSelector
metadata:
  name: test-application
  namespace: default
spec:
  selectors:
  - name: v1beta1 pipelineRef to v1beta1 pipeline
    pipelineRef:
      name: docker-build
      bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:2792f4ae99e98a570a48501f94b727e6c8d60ee6
```

With these changes, build-service will also support v1-style build selectors and v1 pipelines

```yaml
kind: BuildPipelineSelector
metadata:
  name: test-application
  namespace: default
spec:
  selectors:
  - name: v1 pipelineRef to v1beta1 pipeline
    pipelineRef:
      resolver: bundles
      params:
        - name: kind
          value: pipeline
        - name: name
          value: docker-build
        - name: bundle
          value: quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:2792f4ae99e98a570a48501f94b727e6c8d60ee6

  - name: v1beta1 pipelineRef to v1 pipeline
    pipelineRef:
      name: docker-build
      bundle: quay.io/acmiel-test/pipeline-docker-build:tektonapi-v1

  - name: v1 pipelineRef to v1 pipeline
    pipelineRef:
      resolver: bundles
      params:
        - name: kind
          value: pipeline
        - name: name
          value: docker-build
        - name: bundle
          value: quay.io/acmiel-test/pipeline-docker-build:tektonapi-v1
```

More details in commit messages.


chmeliik:
PTAL if this is the approach we want.

Test updates missing, tested only manually on quickcluster

config/crd/patches/fix-tekton-params.yaml
```diff
@@ -1,6 +1,16 @@
...
+      Tekton parameter definition, see https://tekton.dev/docs/pipelines/pipelineruns/#specifying-parameters
+      {
+        name: string,
+        value: string | [string] | {string: string}
+      }
+    type: object
+    # it is not possible to define a structural schema that would allow all the valid values
+    # define params as opaque objects, don't prune any fields
+    # https://kubernetes.io/blog/2019/06/20/crd-structural-schema/#extensions
+    x-kubernetes-preserve-unknown-fields: true
```
chmeliik: This whole commit can be dropped if we want to allow only string values for pipelineRef params

controllers/component_build_controller_pac.go
```diff
@@ -500,12 +500,17 @@ func (r *ComponentBuildReconciler) generatePaCPipelineRunConfigs(ctx context.Con
 	if err != nil {
 		return nil, nil, err
 	}
+	pipelineName, pipelineBundle, err := getNameAndBundle(pipelineRef)
+	if err != nil {
+		// TODO: Log the error? Wrap the error and return it? Record an event?
+		return nil, nil, err
```
chmeliik, line 507: How do we do error handling in build-service?

chmeliik:
I also have no idea how to debug those envtest errors, help would be appreciated :pray:

controllers/component_build_controller_pac.go
tkdchen: 'unsupported Tekton resolver " can be a transient error or normal Go error, that makes the reconciler be rescheduled to give a chance to correct the selector definition.

Refer to perror.go

tkdchen:
Nice. Could you please gather all the information, which is written inside commit messages, into a document for the API migration?

chmeliik:
> I also have no idea how to debug those envtest errors, help would be appreciated 🙏

A lot of the errors are `conversion webhook for tekton.dev/v1, Kind=PipelineRun failed: Post "[https://tekton-pipelines-webhook.tekton-pipelines.svc:443/?timeout=30s](https://tekton-pipelines-webhook.tekton-pipelines.svc/?timeout=30s)": service "tekton-pipelines-webhook" not found`

Maybe the problem is that the tests are creating v1beta1 resources. That would normally work, because Tekton converts those internally, but it sounds like envtest doesn't have the relevant service deployed. Might be fixable by updating the tests to only use v1

controllers/component_build_controller_pac.go
chmeliik: Would the user know that they have to correct the selector definition? Do the error messages for transient errors - or any other errors - somehow get propagated to the user?

chmeliik:
> Maybe the problem is that the tests are creating v1beta1 resources. That would normally work, because Tekton converts those internally, but it sounds like envtest doesn't have the relevant service deployed. Might be fixable by updating the tests to only use v1

So, it was the other way around - tests were creating v1 resources, but kubernetes wanted to store them as v1beta1 because v1beta1 is the storage version for PipelineRuns. Or at least it was, before this commit https://github.com/tektoncd/pipeline/commit/7384a67b77c07444f0e1d5748e771c1477e0db23 (which was released in 0.49.0, but we use 0.44.0)

Fixed by patching the PipelineRun CRD during test suite setup

chmeliik:
Ready for review. Some TODOs left, but I think might your input on some of them

controllers/component_build_controller_pac.go
rcerven: those errors are only in build-service pod logs

difference between transient and non transient (use of perror.go) is that controller either returns error or not
returns error : https://github.com/redhat-appstudio/build-service/blob/main/controllers/component_build_controller.go#L338
that will trigger another reconcile (and repeat few times)

if you don't return error like we do for permanent errors, it will proceed to the end and return nil
https://github.com/redhat-appstudio/build-service/blob/main/controllers/component_build_controller.go#L493
which won't run another reconcile

unsupported tekton resolver seems like persistent error to me

controllers/component_build_controller_simple_build.go
```diff
@@ -83,12 +88,32 @@ func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component
...
 
 	log.Info(fmt.Sprintf("Build pipeline %s created for component %s in %s namespace using %s pipeline from %s bundle",
-		buildPipelineRun.Name, component.Name, component.Namespace, pipelineRef.Name, pipelineRef.Bundle),
+		buildPipelineRun.Name, component.Name, component.Namespace, pipelineName, pipelineBundle),
 		l.Action, l.ActionAdd, l.Audit, "true")
 
 	return nil
 }
 
+func getNameAndBundle(pipelineRef *tektonapi.PipelineRef) (string, string, error) {
```
rcerven, line 97: can we have it in component_build_controller_common.go ?

controllers/component_build_controller_simple_build.go
```diff
@@ -83,12 +88,32 @@ func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component
...
+	if pipelineRef.Resolver != "" && pipelineRef.Resolver != "bundles" {
+		return "", "", fmt.Errorf("unsupported Tekton resolver %q", pipelineRef.Resolver)
```
rcerven, line 99: this should be new error in perror.go (so it won't act like transient error)

controllers/component_build_controller_simple_build.go
```diff
@@ -83,12 +88,32 @@ func (r *ComponentBuildReconciler) SubmitNewBuild(ctx context.Context, component
...
+	}
+
+	name := pipelineRef.Name
+	var bundle string
+
+	for _, param := range pipelineRef.Params {
+		switch param.Name {
+		case "name":
+			name = param.Value.StringVal
+		case "bundle":
```
rcerven, line 109: is it possible not to have 'bundle' specified?

if yes, we  should check it and return error if that is the case

controllers/component_build_controller_test.go
```diff
@@ -1849,3 +1945,13 @@ var _ = Describe("Component initial build controller", func() {
 		})
 	})
 })
+
+func getName(pipelineRef *tektonapi.PipelineRef) string {
```
rcerven, line 1949: getNameFromPRef

or something more explaining

rcerven, line 1949: getNameFromPRef

or something more explaining

controllers/component_build_controller_test.go
```diff
@@ -1849,3 +1945,13 @@ var _ = Describe("Component initial build controller", func() {
...
+	name, _, _ := getNameAndBundle(pipelineRef)
+	return name
+}
+
+func getBundle(pipelineRef *tektonapi.PipelineRef) string {
```
rcerven, line 1954: getBundleFromPRef

or something more explaining

controllers/component_build_controller_pac.go
```diff
@@ -500,12 +501,17 @@ func (r *ComponentBuildReconciler) generatePaCPipelineRunConfigs(ctx context.Con
 	if err != nil {
 		return nil, nil, err
 	}
+	pipelineName, pipelineBundle, err := getNameAndBundle(pipelineRef)
+	if err != nil {
+		// TODO: Log the error? Wrap the error and return it? Record an event?
```
mmorhun, line 506: For this specific case, I think we should just return permanent error, so the error will be visible in the build status annotation (and so on UI).

controllers/component_build_controller_pac.go
mmorhun: It definitely should be a persistent error.

controllers/component_build_controller_pac.go
```diff
@@ -887,18 +893,31 @@ func createWorkspaceBinding(pipelineWorkspaces []tektonapi.PipelineWorkspaceDecl
...
-	if !ok {
+
+	var pipelineSpec tektonapi.PipelineSpec
+
+	if v1beta1Pipeline, ok := obj.(tektonapi_v1beta1.PipelineObject); ok {
+		v1beta1PipelineSpec := v1beta1Pipeline.PipelineSpec()
+		log.Info("Converting from v1beta1 to v1", "PipelineName", pipelineName, "Bundle", bundleUri)
+		err := v1beta1PipelineSpec.ConvertTo(ctx, &pipelineSpec)
+		if err != nil {
+			return nil, fmt.Errorf("pipeline %s from bundle %s: failed to convert from v1beta1 to v1: %w", bundleUri, pipelineName, err)
```
mmorhun, line 914: Looks like a permanent error.

controllers/component_build_controller_simple_build.go
mmorhun, line 97: `getPipelineNameAndBundle` sounds more clear to me.

controllers/component_build_controller_test.go
mmorhun: What about `getPipelineName` ?

go.mod
```diff
@@ -169,7 +169,7 @@ require (
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	k8s.io/apiextensions-apiserver v0.26.1 // indirect
+	k8s.io/apiextensions-apiserver v0.26.1
```
mmorhun, line 172: Direct dependency should not be in indirect dependencies block

controllers/suite_test.go
```diff
@@ -70,12 +75,20 @@ var _ = BeforeSuite(func() {
...
+	Expect(runKustomize(filepath.Join("..", "config", "crd"), crdsTempfile)).To(Succeed())
+	crdsTempfile.Close()
+
 	applicationServiceDepVersion := "v0.0.0-20230717184417-67d31a01a776"
 	applicationApiDepVersion := "v0.0.0-20230704143842-035c661f115f"
 	spiApiDepVersion := "v0.2023.22-0.20230731122342-dfc17adc9829"
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths: []string{
-			filepath.Join("..", "config", "crd", "bases"),
+			crdsTempfile.Name(),
```
mmorhun, line 91: What if `bases` would have another CRD?

pkg/pipeline-selector/pipeline_selector_test.go
```diff
@@ -112,9 +112,17 @@ func TestSelectPipelineForComponent(t *testing.T) {
 						Selectors: []buildappstudiov1alpha1.PipelineSelector{
 							{
 								Name: "Java spring",
-								PipelineRef: tektonapi.PipelineRef{
-									Name:   "java-spring-build-pipeline",
-									Bundle: "java-spring-bundle",
+								PipelineRef: buildappstudiov1alpha1.BackwardsCompatiblePipelineRef{
```
mmorhun, line 115: Does it make sense to have a utility function that takes `Name` and `Bundle` and returns the whole structure? WDYT?

controllers/component_build_controller_simple_build.go
chmeliik: It is possible. Build-service will fail while resolving the bundle, same as it would fail before if a user created a BuildPipelineSelector with no `bundle`

```yaml
pipelineRef:
  name: docker-build
  # bundle: ...
```

controllers/suite_test.go
chmeliik: `kubectl kustomize` will return a yaml file that includes all of them, separated by `---`

Envtest can handle this: https://github.com/kubernetes-sigs/controller-runtime/blob/edb7fe9c520ef2eb5e7d54ded35734bd9a10d71d/pkg/envtest/crd.go#L406 (also tested locally)

controllers/component_build_controller_pac.go
chmeliik: Ack, added a permanent error

controllers/component_build_controller_simple_build.go
chmeliik: makes sense, moved

controllers/component_build_controller_test.go
chmeliik: sounds good :+1:

controllers/component_build_controller_pac.go
chmeliik: Works for me. I'll try to test this in quickcluster to see if the logs will have enough details

controllers/component_build_controller_simple_build.go
chmeliik: renamed :+1:

go.mod
chmeliik: Huh, I thought `go mod tidy` was supposed to handle this. It removed the `// indirect` comment, but didn't move the dependency.

Moved it manually

chmeliik:
Added a permanent error for the `resolver: <something other than "bundles">` case. Still need to test it (and add a unit test for it, probably)

Still need to handle the other permanent error case (converting v1beta1 Pipeline to v1)

pkg/pipeline-selector/pipeline_selector_test.go
chmeliik: That would be good, yeah :+1: 

Where should that function live so that it can be used both in the `controllers/` tests and in `pkg/pipeline-selector/` tests? It looks like you can't have test-only utility functions that would be usable across multiple packages :confused: 

Maybe a public function in `pkg/pipeline-selector/pipeline_selector.go`?

controllers/component_build_controller_simple_build.go
chmeliik: Specifically, the error looks like this and is considered transient
```
Pipelines as Code provision transient error {"controller": "component", "controllerGroup": "appstudio.redhat.com", "controllerKind": "Component", "Component": {"name":"devfile-sample-python-basic","namespace":"default"}, "namespace": "default", "name": "devfile-sample-python-basic", "reconcileID": "2b6c543b-5804-4b92-9cae-14464b74f2b4", "error": " is an unparseable image reference: could not parse reference: "}
```
I can add a permanent error for this too

controllers/component_build_controller_pac.go
chmeliik: They do, nice.

```
Pipelines as Code provision for the Component failed {"controller": "component", "controllerGroup": "appstudio.redhat.com", "controllerKind": "Component", "Component": {"name":"devfile-sample-python-basic","namespace":"default"}, "namespace": "default", "name": "devfile-sample-python-basic", "reconcileID": "3c66bec3-5ec2-4736-9630-cb259eb93d7c", "error": "unsupported Tekton resolver \"git\""}
```
```
simple build submition for the Component failed {"controller": "component", "controllerGroup": "appstudio.redhat.com", "controllerKind": "Component", "Component": {"name":"devfile-sample-python-basic","namespace":"default"}, "namespace": "default", "name": "devfile-sample-python-basic", "reconcileID": "d590f08d-8187-46c6-9fe5-a1b118274fe0", "error": "unsupported Tekton resolver \"git\""}
```

controllers/component_build_controller_pac.go
chmeliik: Added a new permanent error for this, PTAL at the message and error ID (not sure what rules those should follow)

chmeliik:
Added tests for the unsupported resolver case, added a new permanent error for failing to convert a v1beta1 Pipeline

controllers/component_build_controller_pac.go
```diff
@@ -887,18 +892,34 @@ func createWorkspaceBinding(pipelineWorkspaces []tektonapi.PipelineWorkspaceDecl
...
+		if err != nil {
+			return nil, boerrors.NewBuildOpError(
+				boerrors.EPipelineConversionFailed,
+				fmt.Errorf("pipeline %s from bundle %s: failed to convert from v1beta1 to v1: %w", pipelineName, bundleUri, err),
+			)
+		}
+	} else if v1Pipeline, ok := obj.(*tektonapi.Pipeline); ok {
+		pipelineSpec = v1Pipeline.PipelineSpec()
+	} else {
+		return nil, fmt.Errorf("failed to extract pipeline %s from bundle %s", pipelineName, bundleUri)
```
mmorhun, line 921: This is permanent error as well.

pkg/pipeline-selector/pipeline_selector_test.go
mmorhun: It mostly used in tests, so I'd put it a test file.

chmeliik: That's the thing, I defined it in `pkg/pipeline-selector/pipeline_selector_test.go`, but now I can't import it in the controller tests
image

According to Reddit, it's not really possible to share test-only utilities across packages https://www.reddit.com/r/golang/comments/jjmsoc/how_to_share_test_utilities_accros_packages/

I guess I can copy-paste the helper function in both `pkg/pipeline-selector` tests and `controllers` tests

controllers/component_build_controller_pac.go
chmeliik: Ack, added a new one.

Don't have a test for it though :disappointed:
I would need a bundle image that contains a pipeline with a version other than v1beta1 or v1

controllers/component_build_controller_simple_build.go
rcerven: yeah in that case I'd return permanent error in that function when bundle is empty

controllers/component_build_controller_common.go
```diff
@@ -305,3 +305,26 @@ func getPathContext(gitContext, dockerfileContext string) string {
...
+
+	for _, param := range pipelineRef.Params {
+		switch param.Name {
+		case "name":
+			name = param.Value.StringVal
+		case "bundle":
+			bundle = param.Value.StringVal
+		}
+	}
+
```
rcerven, line 328: since bundle can be empty

I'd return permanent error here when bundle is empty

controllers/suite_test.go
rcerven: well, there are now two yamls in bases, and 1 kustomize which changes both of them

but if there will be new yaml in bases, without change in kustomize I'd say that :
filepath.Join("..", "config", "crd", "bases"),

would be still required

so I'd just keep it there

rcerven (changes requested):

controllers/suite_test.go
chmeliik: By my understanding, if a second CRD were added we would want to add it to the kustomization.yaml `resources` anyway. The `make install` command step makes the same assumption - `kustomize build config/crd` will return all the CRDs
```make
.PHONY: install
install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
    $(KUSTOMIZE) build config/crd | kubectl apply -f - || true
```

chmeliik:
Dropped the commit that allows any param values in BuildPipelineSelector pipelineRefs, now only strings are allowed

controllers/suite_util_test.go
```diff
@@ -60,8 +60,10 @@ const (
 	ComponentContainerImage = "registry.io/username/image:tag"
 	SelectorDefaultName     = "default"
 
-	defaultPipelineName   = "docker-build"
-	defaultPipelineBundle = "quay.io/redhat-appstudio-tekton-catalog/pipeline-docker-build:8cf8982d58a841922b687b7166f0cfdc1cc3fc72"
+	defaultPipelineName = "docker-build"
+	// v1 Pipeline, from https://github.com/redhat-appstudio/build-definitions/pull/656
+	defaultPipelineBundle = "quay.io/redhat-appstudio-tekton-catalog/pull-request-builds:docker-build-97a993a5e19217381cdb9d1fd3df57a1647afd9a"
```
chmeliik, line 65: Hmm, wait, do `pull-request-builds` get garbage collected?

mmorhun: Yes, by default within 5 days

mmorhun (approved):
LGTM

chmeliik:
Squashed the temp commit (the helper function that creates a pipelineRef with a bundle resolver) into the appropriate commits. Overall no change

rcerven (approved):

PR #210, opened 2023-11-03
STONEBLD-1844 Add controller for PipelineRuns

stuartwdouglas:
This controller looks for build PipelineRuns and will record an event that a downstream component should be 'nudged'.

This is the first part of STONEBLD-1756.

### DIFF
--- a/controllers/component_dependency_update_controller.go
+++ b/controllers/component_dependency_update_controller.go
@@ -1,0 +1,511 @@
+/*
+Copyright 2021-2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	applicationapi "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	l "github.com/redhat-appstudio/build-service/pkg/logs"
+	tektonapi "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	"golang.org/x/exp/slices"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"knative.dev/pkg/apis"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"strconv"
+	"strings"
+	"text/template"
+	"time"
+)
+
+const (
+	contextTimeout = 300 * time.Second
+	// PipelineRunTypeLabelName contains the type of the PipelineRunType
+	PipelineRunTypeLabelName = "pipelines.appstudio.openshift.io/type"
+	// PipelineRunBuildType is the type denoting a build PipelineRun.
+	PipelineRunBuildType = "build"
+	// PacEventTypeAnnotationName represents the current event type
+	PacEventTypeAnnotationName = "pipelinesascode.tekton.dev/event-type"
+	PacEventPushType           = "push"
+	ImageUrlParamName          = "IMAGE_URL"
+	ImageDigestParamName       = "IMAGE_DIGEST"
+
+	NudgeProcessedAnnotationName = "build.appstudio.openshift.io/component-nudge-processed"
+	NudgeFinalizer               = "build.appstudio.openshift.io/build-nudge-finalizer"
+	FailureCountAnnotationName   = "build.appstudio.openshift.io/build-nudge-failures"
+
+	ComponentNudgedEventType      = "ComponentNudged"
+	ComponentNudgeFailedEventType = "ComponentNudgeFailed"
+	MaxAttempts                   = 3
+	KubeApiUpdateMaxAttempts      = 5
+
+	FailureRetryTime = time.Minute * 5 // We retry after 5 minutes on failure
+)
+
+// ComponentDependencyUpdateReconciler reconciles a PipelineRun object
+type ComponentDependencyUpdateReconciler struct {
+	client.Client
+	ApiReader      client.Reader
+	Scheme         *runtime.Scheme
+	EventRecorder  record.EventRecorder
+	UpdateFunction UpdateComponentDependenciesFunction
+}
+
+type BuildResult struct {
+	BuiltImageRepository     string
+	BuiltImageTag            string
+	Digest                   string
+	DistributionRepositories []string
+	Component                *applicationapi.Component
+}
+
+// SetupController creates a new Integration reconciler and adds it to the Manager.
+func (r *ComponentDependencyUpdateReconciler) SetupWithManager(manager ctrl.Manager) error {
+	return setupControllerWithManager(manager, r)
+}
+
+// setupControllerWithManager sets up the controller with the Manager which monitors new PipelineRuns and filters
+// out pipelines we don't need
+func setupControllerWithManager(manager ctrl.Manager, reconciler *ComponentDependencyUpdateReconciler) error {
+	return ctrl.NewControllerManagedBy(manager).
+		For(&tektonapi.PipelineRun{}, builder.WithPredicates(predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool {
+				new, ok := e.Object.(*tektonapi.PipelineRun)
+				if !ok {
+					return false
+				}
+				return IsBuildPushPipelineRun(new)
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				new, ok := e.ObjectNew.(*tektonapi.PipelineRun)
+				if !ok {
+					return false
+				}
+				if !IsBuildPushPipelineRun(new) {
+					return false
+				}
+
+				// Ensure we have not finished processing
+				if new.ObjectMeta.Annotations != nil && new.ObjectMeta.Annotations[NudgeProcessedAnnotationName] != "" {
+					return false
+				}
+				return true
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return true
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				return true
+			},
+		})).
+		Complete(reconciler)
+}
+
+type UpdateComponentDependenciesFunction = func(ctx context.Context, client client.Client, scheme *runtime.Scheme, eventRecorder record.EventRecorder, downstreamComponents []applicationapi.Component, result *BuildResult) (immediateRetry bool, err error)
+
+var DefaultUpdateFunction = DefaultDependenciesUpdate
+
+// +kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns,verbs=get;list;watch;create;update;patch;delete;deletecollection
+// +kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+func (r *ComponentDependencyUpdateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
+	log := ctrllog.FromContext(ctx).WithName("ComponentNudge")
+	ctx = ctrllog.IntoContext(ctx, log)
+
+	pipelineRun := &tektonapi.PipelineRun{}
+	err := r.Get(ctx, req.NamespacedName, pipelineRun)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		log.Error(err, "Failed to get pipelineRun")
+		return ctrl.Result{}, err
+	}
+
+	component, err := GetComponentFromPipelineRun(r.Client, ctx, pipelineRun)
+	if err != nil || component == nil {
+		log.Error(err, "failed to get component")
+		return ctrl.Result{}, err
+	}
+
+	if len(component.Spec.BuildNudgesRef) == 0 {
+		log.Info(fmt.Sprintf("component %s has no BuildNudgesRef set", component.Name))
+		return ctrl.Result{}, nil
+	}
+	log.Info("reconciling PipelineRun")
+
+	if pipelineRun.IsDone() || pipelineRun.Status.CompletionTime != nil || pipelineRun.DeletionTimestamp != nil {
+		log.Info("PipelineRun complete")
+		// These objects are so heavily contented that we always grab the latest copy from the
+		// API server to reduce the chance of conflicts
+		err = r.ApiReader.Get(ctx, req.NamespacedName, pipelineRun)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if controllerutil.ContainsFinalizer(pipelineRun, NudgeFinalizer) || pipelineRun.Annotations == nil || pipelineRun.Annotations[NudgeProcessedAnnotationName] == "" {
+			log.Info("running renovate job")
+			// Pipeline run is done and we have not cleared the finalizer yet
+			// We need to perform our nudge
+			return r.handleCompletedBuild(ctx, pipelineRun, component)
+		}
+	} else if !controllerutil.ContainsFinalizer(pipelineRun, NudgeFinalizer) {
+		log.Info("adding finalizer for component nudge")
+		// These objects are so heavily contented that we always grab the latest copy from the
+		// API server to reduce the chance of conflicts
+		err := r.ApiReader.Get(ctx, req.NamespacedName, pipelineRun)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		// We add a finalizer to make sure we see the run before it is deleted
+		// As tekton results should aggressivly delete when pruning is enabled
+		controllerutil.AddFinalizer(pipelineRun, NudgeFinalizer)
+		err = r.Client.Update(ctx, pipelineRun)
+		if err == nil {
+			return reconcile.Result{}, nil
+		} else if errors.IsConflict(err) {
+			log.Error(err, "failed to add finalizer due to conflict, retrying in one second")
+			return reconcile.Result{RequeueAfter: time.Second}, nil
+		} else {
+			log.Error(err, "failed to add finalizer")
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+// handleCompletedBuild will perform a 'nudge' updating dependent downstream components.
+// This will involve creating a PR updating their references to our images to the newly produced image
+func (r *ComponentDependencyUpdateReconciler) handleCompletedBuild(ctx context.Context, pipelineRun *tektonapi.PipelineRun, updatedComponent *applicationapi.Component) (ctrl.Result, error) {
+	log := ctrllog.FromContext(ctx)
+	success := pipelineRun.Status.GetCondition(apis.ConditionSucceeded).IsTrue()
+	if !success {
+		log.Info("not performing nudge as pipeline failed")
+		return r.removePipelineFinalizer(ctx, pipelineRun)
+	}
+	// find the image and digest we want to update to
+	image := ""
+	digest := ""
+	for _, r := range pipelineRun.Status.Results {
+		if r.Name == ImageDigestParamName {
+			digest = r.Value.StringVal
+		} else if r.Name == ImageUrlParamName {
+			image = r.Value.StringVal
+		}
+	}
+	// Failure to find is a permanent error, so remove the finalizer
+	if image == "" {
+		log.Error(fmt.Errorf("unable to find %s param on PipelineRun, not performing nudge", ImageUrlParamName), "no image url result")
+		return r.removePipelineFinalizer(ctx, pipelineRun)
+	}
+	if digest == "" {
+		log.Error(fmt.Errorf("unable to find %s param on PipelineRun, not performing nudge", ImageDigestParamName), "no image digest result")
+		return r.removePipelineFinalizer(ctx, pipelineRun)
+	}
+	tag := ""
+	repo := image
+	index := strings.LastIndex(image, ":")
+	if index != -1 {
+		repo = image[0:index]
+		tag = image[index+1:]
+	}
+
+	components := applicationapi.ComponentList{}
+	err := r.Client.List(ctx, &components, client.InNamespace(pipelineRun.Namespace))
+	if err != nil {
+		log.Error(err, "failed to list components in namespace")
+		return ctrl.Result{}, err
+	}
+
+	retryTime := FailureRetryTime
+	immediateRetry := false
+
+	toUpdate := []applicationapi.Component{}
+
+	componentDesc := ""
+	for i := range components.Items {
+		comp := components.Items[i]
+		if comp.Spec.Application == updatedComponent.Spec.Application && slices.Contains(updatedComponent.Spec.BuildNudgesRef, comp.Name) {
+			toUpdate = append(toUpdate, comp)
+		}
+		if componentDesc != "" {
+			componentDesc += ", "
+		}
+		componentDesc += comp.Namespace + "/" + comp.Name
+	}
+	var nudgeErr error
+	immediateRetry, nudgeErr = r.UpdateFunction(ctx, r.Client, r.Scheme, r.EventRecorder, toUpdate, &BuildResult{BuiltImageRepository: repo, BuiltImageTag: tag, Digest: digest, Component: updatedComponent})
+
+	if nudgeErr != nil {
+
+		// These objects are so heavily contented that we always grab the latest copy from the
+		// API server to reduce the chance of conflicts
+		err = r.ApiReader.Get(ctx, types.NamespacedName{Namespace: pipelineRun.Namespace, Name: pipelineRun.Name}, pipelineRun)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		log.Error(nudgeErr, fmt.Sprintf("component update of components %s as a result of a build of %s failed", componentDesc, updatedComponent.Name), l.Audit, "true")
+
+		if pipelineRun.Annotations == nil {
+			pipelineRun.Annotations = map[string]string{}
+		}
+		existing := pipelineRun.Annotations[FailureCountAnnotationName]
+		if existing == "" {
+			existing = "0"
+		}
+		failureCount, err := strconv.Atoi(existing)
+		if err != nil {
+			log.Error(err, "failed to parse retry count, not retrying")
+			return r.removePipelineFinalizer(ctx, pipelineRun)
+		}
+		failureCount = failureCount + 1
+
+		pipelineRun.Annotations[FailureCountAnnotationName] = strconv.Itoa(failureCount)
+
+		r.EventRecorder.Event(updatedComponent, corev1.EventTypeWarning, ComponentNudgeFailedEventType, fmt.Sprintf("component update failed as a result of a build for %s, retry %d/%d", updatedComponent.Name, failureCount, MaxAttempts))
+
+		if failureCount >= MaxAttempts {
+			// We are at the failure limit, nothing much we can do
+			log.Info("not retrying as max failure limit has been reached", l.Audit, "true")
+			return r.removePipelineFinalizer(ctx, pipelineRun)
+		}
+		log.Info(fmt.Sprintf("failed to update component dependencies, retry %d/%d", failureCount, MaxAttempts))
+		err = r.Client.Update(ctx, pipelineRun)
+		if err != nil {
+			// If we fail to update just return and let requeue handle it
+			// We can't really do anything else
+			// This does mean components may get nudged a second time, but it is idempotent anyway
+			return ctrl.Result{}, err
+		}
+		if immediateRetry {
+			return reconcile.Result{RequeueAfter: time.Millisecond}, nil
+		} else {
+			return reconcile.Result{RequeueAfter: retryTime * time.Duration(failureCount)}, nil
+		}
+	}
+
+	_, err = r.removePipelineFinalizer(ctx, pipelineRun)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Now we need to look for 'stale' pipelines.
+	// These are defined as pipelines that are younger than this one, that target the same component.
+	// If there are two pushes very close together variances in pipeline run times may mean that the
+	// older one finishes last. In this case we want to mark the older one as already nudged and remove
+	// the finalizer.
+
+	pipelines := tektonapi.PipelineRunList{}
+	err = r.Client.List(ctx, &pipelines, client.InNamespace(pipelineRun.Namespace), client.MatchingLabels{PipelineRunTypeLabelName: PipelineRunBuildType, ComponentNameLabelName: updatedComponent.Name})
+	if err != nil {
+		// I don't think we want to retry this, it should be really rare anyway
+		// and would require an even more complex label based state machine.
+		log.Error(err, "failed to check for stale pipeline runs, this operation will not be retried", l.Audit, "true")
+		return ctrl.Result{}, nil
+	}
+	var finalizerError error
+	for i := range pipelines.Items {
+		possiblyStalePr := pipelines.Items[i]
+		if possiblyStalePr.Annotations == nil || possiblyStalePr.Annotations[PacEventTypeAnnotationName] != PacEventPushType || possiblyStalePr.Name == pipelineRun.Name {
+			continue
+		}
+		if possiblyStalePr.Status.CompletionTime == nil && possiblyStalePr.CreationTimestamp.Before(&pipelineRun.CreationTimestamp) {
+			log.Info(fmt.Sprintf("marking PipelineRun %s as nudged, as it is stale", possiblyStalePr.Name))
+			if possiblyStalePr.Annotations == nil {
+				possiblyStalePr.Annotations = map[string]string{}
+			}
+			_, err := r.removePipelineFinalizer(ctx, &possiblyStalePr)
+			if err != nil {
+				finalizerError = err
+				log.Error(err, "failed to update stale pipeline run", l.Audit, "true")
+			}
+		}
+
+	}
+	return ctrl.Result{}, finalizerError
+}
+
+// removePipelineFinalizer will remove the finalizer, and add an annotation to indicate we are done with this pipeline run
+// We can't use just the presence or absence of the finalizer, as there is some situations where we might not have seen the
+// run until it is completed, e.g. if the controller was down.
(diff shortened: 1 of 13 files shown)


controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		log.Error(err, "Failed to get pipelineRun")
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, err
+	}
+
+	component, err := GetComponentFromPipelineRun(r.Client, ctx, pipelineRun)
```
mmorhun, line 91: `component == nil` not handled

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		log.Error(err, "failed to get component")
+		return ctrl.Result{}, err
+	}
+
+	if len(component.Spec.BuildNudgesRef) == 0 {
+		log.Info(fmt.Sprintf("component %s has no BuildNudgesRef set", component.Name))
+		return ctrl.Result{}, err
```
mmorhun, line 99: Why do we need to return `err` here instead of `nil`?

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	if pipelineRun.IsDone() || pipelineRun.DeletionTimestamp != nil {
```
mmorhun, line 102: Pipeline can be deleted before completion.
Wouldn't it be better to analyze `DeletionTimestamp` at the beginning of the reconcile?

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+			//we need to perform our nudge
+			return r.handleCompletedBuild(ctx, log, pipelineRun, component)
+		}
+	} else if !controllerutil.ContainsFinalizer(pipelineRun, NudgeFinalizer) {
+		//we add a finalizer to make sure we see the run before it is deleted
+		//as tekton results should aggressivly delete when pruning is enabled
+		controllerutil.AddFinalizer(pipelineRun, NudgeFinalizer)
+		err := r.Client.Update(ctx, pipelineRun)
+		if err != nil {
+			log.Error(err, "failed to add finalizer")
```
mmorhun, line 114: Shouldn't we retry?

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	if len(component.Spec.BuildNudgesRef) == 0 {
+		log.Info(fmt.Sprintf("component %s has no BuildNudgesRef set", component.Name))
+		return ctrl.Result{}, err
+	}
+
+	if pipelineRun.IsDone() || pipelineRun.DeletionTimestamp != nil {
+		if controllerutil.ContainsFinalizer(pipelineRun, NudgeFinalizer) || pipelineRun.Annotations == nil || pipelineRun.Annotations[NudgeProcessedAnnotationName] == "" {
+			//pipeline run is done and we have not cleared the finalizer yet
```
mmorhun, line 104: nitpick: other controller use space after `//` and start comment from capital letter.

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+const (
+	contextTimeout = 300 * time.Second
+	// PipelineRunTypeLabel contains the type of the PipelineRunTypeLabel.
+	PipelineRunTypeLabel = "pipelines.appstudio.openshift.io/type"
+	// PipelineRunBuildType is the type denoting a build PipelineRun.
+	PipelineRunBuildType         = "build"
+	NudgeProcessedAnnotationName = "build.appstudio.openshift.io/component-nudge-processed"
+
+	// PipelineRunComponentLabel is the label denoting the application.
+	PipelineRunComponentLabel = "appstudio.openshift.io/component"
```
mmorhun, line 50: nitpick: other controller and annotation above use `Name` suffix to distinguish label/key etc name from the value constant.

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	NudgeFinalizer            = "appstudio.openshift.io/build-nudge-finalizer"
+
+	PacEventType     = "pipelinesascode.tekton.dev/event-type"
```
mmorhun, line 53: Is it label name?

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+const (
+	contextTimeout = 300 * time.Second
+	// PipelineRunTypeLabel contains the type of the PipelineRunTypeLabel.
+	PipelineRunTypeLabel = "pipelines.appstudio.openshift.io/type"
+	// PipelineRunBuildType is the type denoting a build PipelineRun.
+	PipelineRunBuildType         = "build"
+	NudgeProcessedAnnotationName = "build.appstudio.openshift.io/component-nudge-processed"
```
mmorhun, line 47: What's teh purpose of this annotation? Cannot we use finalyzer to figure out the same?

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+// +kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns,verbs=get;list;watch;create;update;patch;delete;deletecollection
+// +kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+func (r *ComponentNudgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithTimeout(ctx, contextTimeout)
+	defer cancel()
```
mmorhun, line 76: Could you explain what could be ongoing we want to stop when the reconcile ends?

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	if digest == "" {
+		log.Error(fmt.Errorf("unable to find %s param on PipelineRun, not performing nudge", ImageDigestParam), "no image digest result")
+		return ctrl.Result{}, nil
+	}
+
+	for i := range components.Items {
+		comp := components.Items[i]
+		if comp.Spec.Application == updatedComponent.Spec.Application && slices.Contains(updatedComponent.Spec.BuildNudgesRef, comp.Name) {
+			//TODO: Do the nudge
```
mmorhun, line 167: Nudge would be a complicated operation that takes some time and involves dealing with git provider API.
What if nudge for one of the components fails? What if it's transient/consistent error?

main.go
```diff
@@ -164,6 +164,24 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err = (&controllers.GitTektonResourcesRenovater{
```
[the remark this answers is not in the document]
stuartwdouglas: Oops, I stared looking at the next steps and must have accidentally committed some of my experiments. I was just trying to do too many things at once yesterday.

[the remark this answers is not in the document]
stuartwdouglas: Fixed

controllers/component_pipelinerun_nudge_controller.go
```diff
@@ -0,0 +1,220 @@
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// GetComponentFromPipelineRun loads from the cluster the Component referenced in the given PipelineRun. If the PipelineRun doesn't
+// specify a Component or this is not found in the cluster, an error will be returned.
+func GetComponentFromPipelineRun(c client.Client, ctx context.Context, pipelineRun *tektonapi.PipelineRun) (*applicationapiv1alpha1.Component, error) {
```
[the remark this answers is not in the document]
stuartwdouglas: Kinda, but we need the finalizer as well to prevent tekton results pruning the pipeline before we have reacted to it (if tekton results pruning ever gets enabled).

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: copy/paste error, err would always be nil here anyway.

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: We need to remove the finalizer no matter what if the PR is deleted, deleted before completion is handled in the handleCompletedBuild as it will not be considered successful.

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: What if the pipeline run completes before we added the finalize (e.g. if the controller is down for a little bit)? Basically we can't be sure that we will reconcile the pipeline run before it completes.

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: I thought this was just good practice for operator code. At the moment there is nothing, but the lack of this was actually the cause of the tekton results lockups we were experiencing a while back. I can delete it until we actually need it if you want.

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: This story was just about getting it to the point where we can print the message in the logs. I assume we will need a retry count mechanism with back off, we don't want to retry indefinitely, but we also don't want a single transient failure to stop the update.

stuartwdouglas:
@mmorhun I think I have addressed everything

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,220 @@
...
+		err := r.Client.Update(ctx, pipelineRun)
+		if err != nil {
+			log.Error(err, "failed to add finalizer")
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *ComponentDependencyUpdateReconciler) handleCompletedBuild(ctx context.Context, log logr.Logger, pipelineRun *tektonapi.PipelineRun, updatedComponent *applicationapiv1alpha1.Component) (ctrl.Result, error) {
```
tkdchen, line 135: Could you please comment this function?

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,224 @@
...
+	// Note that we do this eagerly, as we only really have one chance to do the nudge
+	// We don't want to get stuck in a nudge loop, or fail to remove the finalizer
+	if pipelineRun.Annotations == nil {
+		pipelineRun.Annotations[NudgeProcessedAnnotationName] = "true"
+	}
+	controllerutil.RemoveFinalizer(pipelineRun, NudgeFinalizer)
+	err := r.Client.Update(ctx, pipelineRun)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
tkdchen, line 149: This may determine any failure nudge will be a permanent error and the retry will not depend on reconcile, it needs another PipelineRun triggered again. Is my understanding correct?

tkdchen, line 149: Would be good to emit an Event for the failure of removing the finalizer.

stuartwdouglas: See this comment: https://github.com/redhat-appstudio/build-service/pull/210#discussion_r1382294356

This story was just about the initial reconcile loop and logging a message.

stuartwdouglas:
Given that both @tkdchen and @mmorhun asked about failure handling I have added failure handling and the retry mechanism to this PR. Currently it won't really do anything as printing log messages won't fail, but there are unit tests for artificial failure.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+func (r *ComponentDependencyUpdateReconciler) handleCompletedBuild(ctx context.Context, log logr.Logger, pipelineRun *tektonapi.PipelineRun, updatedComponent *applicationapiv1alpha1.Component) (ctrl.Result, error) {
+
+	success := pipelineRun.Status.GetCondition(apis.ConditionSucceeded).IsTrue()
+	if !success {
+		log.Info("not performing nudge as pipeline failed")
+		return r.removePipelineFinalizer(ctx, log, pipelineRun)
+	}
+
+	components := applicationapiv1alpha1.ComponentList{}
+	err := r.Client.List(ctx, &components, client.InNamespace(pipelineRun.Namespace))
```
mmorhun, line 156: I think this block should be moved after image/digest block below.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+			return ctrl.Result{}, err
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+// handleCompletedBuild will perform a 'nudge' updating dependent downstream components
+// this will involve creating a PR updating their references to our images to the newly produced image
+// for now it just produces a log message
+func (r *ComponentDependencyUpdateReconciler) handleCompletedBuild(ctx context.Context, log logr.Logger, pipelineRun *tektonapi.PipelineRun, updatedComponent *applicationapiv1alpha1.Component) (ctrl.Result, error) {
```
mmorhun, line 147: We can store logger in the context, see other controllers.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+	NudgeFinalizer                = "appstudio.openshift.io/build-nudge-finalizer"
+
+	PacEventTypeLabelName = "pipelinesascode.tekton.dev/event-type"
+	PacEventPush          = "push"
+	ComponentNudged       = "ComponentNudged"
+	ComponentNudgeFailed  = "ComponentNudgeFailed"
+	ImageUrlParam         = "IMAGE_URL"
+	ImageDigestParam      = "IMAGE_DIGEST"
+
+	TestFailureAnnotation  = "internal.testing.appstudio.openshift.io/unit-test-failure-count"
```
mmorhun, line 62: Having an annotation for testing looks like overhead. Also, it's possible to influence the flow using it.
I'd suggest moving nudge for component into a separate method/function and mock it in tests like we do for git clients, for example.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+		if pipelineRun.Annotations == nil {
+			pipelineRun.Annotations = map[string]string{}
+		}
+		existing := pipelineRun.Annotations[FailureCountAnnotation]
+		if existing == "" {
+			existing = "0"
+		}
+		failureCount, err := strconv.Atoi(existing)
+		if err != nil {
+			log.Error(err, "failed to pass retry count, not retrying")
```
mmorhun, line 215: `pass`? did you mean `parse`?

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+			return r.removePipelineFinalizer(ctx, log, pipelineRun)
+		}
+		failureCount = failureCount + 1
+
+		r.EventRecorder.Event(updatedComponent, v12.EventTypeWarning, ComponentNudgeFailed, fmt.Sprintf("component update failed as a result of a build for %s, retry %d/%d", updatedComponent.Name, failureCount, MaxAttempts))
+		log.Error(nudgeErr, "component update failed")
```
mmorhun, line 221: Could you mark this log as audit log entry?

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+		pipelineRun.Annotations[FailureCountAnnotation] = strconv.Itoa(failureCount)
+		if failureCount >= MaxAttempts {
+			//we are at the failure limit, nothing much we can do
+			log.Error(nudgeErr, "not retrying as max failure limit has been reached")
```
mmorhun, line 225: Could you mark this log as audit log entry?

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+	PacEventPush          = "push"
+	ComponentNudged       = "ComponentNudged"
+	ComponentNudgeFailed  = "ComponentNudgeFailed"
+	ImageUrlParam         = "IMAGE_URL"
+	ImageDigestParam      = "IMAGE_DIGEST"
+
+	TestFailureAnnotation  = "internal.testing.appstudio.openshift.io/unit-test-failure-count"
+	FailureCountAnnotation = "appstudio.openshift.io/build-nudge-failures"
+	MaxAttempts            = 3
+	FailureRetryTime       = time.Minute * 5 // We retry after 5 minutes on failure
```
mmorhun, line 65: Not sure that we should wait 5*x minutes after a failure. It could be a network error as well.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	components := applicationapiv1alpha1.ComponentList{}
+	err := r.Client.List(ctx, &components, client.InNamespace(pipelineRun.Namespace))
+	if err != nil {
+		log.Error(err, "failed to list components in namespace")
+		return ctrl.Result{}, err
+	}
+
+	//find the image and digest we want to update to
```
mmorhun, line 162: nitpick: please align comments style (here and below)

go.mod
```diff
@@ -176,8 +176,8 @@ require (
 	k8s.io/klog v1.0.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20230501164219-8b0f38b5fd1f // indirect
 	k8s.io/pod-security-admission v0.26.1 // indirect
-	k8s.io/utils v0.0.0-20230209194617-a36077c30491 // indirect
-	knative.dev/pkg v0.0.0-20230125083639-408ad0773f47 // indirect
+	k8s.io/utils v0.0.0-20230209194617-a36077c30491
+	knative.dev/pkg v0.0.0-20230125083639-408ad0773f47
```
mmorhun, line 180: Please move all direct dependencies into corresponding `require` block.

controllers/component_pipelinerun_nudge_controller.go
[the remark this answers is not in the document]
mmorhun: How does the predicate influence the finalyzer?

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: We will need to add an annotation with the state information. Also nudging should be idempotent, if you run the same nudge again it should figure out that there is an existing PR, and the update to the PR would result in the same commit.

stuartwdouglas: I have added support for this.

go.mod
```diff
@@ -149,7 +149,7 @@ require (
 	go.uber.org/multierr v1.9.0 // indirect
 	go.uber.org/zap v1.24.0
 	golang.org/x/crypto v0.11.0 // indirect
-	golang.org/x/exp v0.0.0-20230206171751-46f607a40771 // indirect
+	golang.org/x/exp v0.0.0-20230206171751-46f607a40771
```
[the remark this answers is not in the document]
stuartwdouglas: I thought go mod tidy was supposed to handle this automatically, at least in the JVM build service it does. It must not handle the case when an indirect is changed to a direct.

Some of these indirect -> direct changes are not actually a result of this PR, and just a result of running go mod tidy, and there are some existing deps that are direct and in the indirect section.

controllers/component_pipelinerun_nudge_controller.go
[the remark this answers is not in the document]
stuartwdouglas: It will just need to pass through more events, as it also needs to see new runs.

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: If the annotation is present we know the pipeline run has been processed. Without it the lack of a finalizer we can't really distinguish between 'an object we have not seen before', and 'an object that is fully processed'. 

Basically if this was down for a few minutes and then we see a completed pipeline run with no finalizer on startup, we can't know if this was a run that we have never seen before that completed while we were down, or a run that has already had its dependencies processed.

controllers/component_pipelinerun_nudge_controller.go
stuartwdouglas: Something to bear in mind with error handling too is that the actual updates will be async, and error handling will also need to be done through watching the job/pipeline that is doing the update.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,373 @@
...
+	FailureRetryTime              = time.Minute * 5 // We retry after 5 minutes on failure
+)
+
+// ComponentDependencyUpdateReconciler reconciles a PipelineRun object
+type ComponentDependencyUpdateReconciler struct {
+	client.Client
+	Scheme         *runtime.Scheme
+	EventRecorder  record.EventRecorder
+	UpdateFunction UpdateComponentDependenciesFunction
+}
```
tkdchen, line 81: What's the consideration for adding a function pointer `UpdateFunction`? Easy for writing tests?

stuartwdouglas: To allow it to be easily mocked in tests.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,373 @@
...
+
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return controllerutil.ContainsFinalizer(e.Object, NudgeFinalizer)
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				return false
+			},
+		})).
+		WithEventFilter(predicate.NewPredicateFuncs(IsBuildPushPipelineRun)).
```
tkdchen, line 132: How about removing this call and applying `IsBuildPushPipelineRun` in the newly added predicates?

stuartwdouglas: ah, we don't need that now that we have the predicate.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,375 @@
...
+		//if any
+		if len(alreadyProcessed) > 0 {
+			val := ""
+			for i := range alreadyProcessed {
+				if val != "" {
+					val = val + ","
+				}
+				val = val + i
+			}
+			pipelineRun.Annotations[AlreadyNudgedAnnotationName] = val
```
mmorhun, line 282: What if a new build will start in meantime?

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,375 @@
...
+		}
+		failureCount = failureCount + 1
+
+		pipelineRun.Annotations[FailureCountAnnotationName] = strconv.Itoa(failureCount)
+
+		//update the list of successful updates
+		//if any
+		if len(alreadyProcessed) > 0 {
+			val := ""
+			for i := range alreadyProcessed {
```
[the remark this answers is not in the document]
stuartwdouglas: It's a map not a slice. I could use maps.Keys() to get a slice and call join on that I suppose.

controllers/component_dependency_update_controller.go
stuartwdouglas: That is what I mentioned above about builds going stale. When I perform the 'nudge' I could check for currently running pipelines for the same component that are older than the current state, and if they exist and are still running mark them as already nudged and remove the finalizer.

The other approach would be that when a pipeline finishes we check if a newer pipeline exists and has finished, however I don't think that approach will work very well as the pruner may have removed already completed pipelines.

stuartwdouglas: I have implemented the stale check logic

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,407 @@
...
+	err = r.Client.List(ctx, &pipelines, client.InNamespace(pipelineRun.Namespace), client.MatchingLabels{PipelineRunTypeLabelName: PipelineRunBuildType, PipelineRunComponentLabelName: updatedComponent.Name})
+	if err != nil {
+		// I don't think we want to retry this, it should be really rare anyway
+		// And would require an even more complex label based state machine
+		log.Error(err, "failed to check for stale pipeline runs, this operation will not be retried", l.Audit, "true")
+		return ctrl.Result{}, nil
+	}
+	for i := range pipelines.Items {
+		possiblyStalePr := pipelines.Items[i]
+		if possiblyStalePr.Annotations == nil || possiblyStalePr.Annotations[PacEventTypeAnnotationName] != PacEventPushType || possiblyStalePr.Name == pipelineRun.Name {
```
mmorhun, line 322: Should we also skip pipelines with `NudgeProcessedAnnotationName` annotation?

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,407 @@
...
+		}
+		if possiblyStalePr.Status.CompletionTime == nil && possiblyStalePr.CreationTimestamp.Before(&pipelineRun.CreationTimestamp) {
+			log.Info(fmt.Sprintf("marking PipelineRun %s as nudged, as it is stale", possiblyStalePr.Name))
+			if possiblyStalePr.Annotations == nil {
+				possiblyStalePr.Annotations = map[string]string{}
+			}
+			possiblyStalePr.Annotations[NudgeProcessedAnnotationName] = "true"
+			_, err := r.removePipelineFinalizer(ctx, &possiblyStalePr)
+			if err != nil {
+				//TODO: Do we need a retry here on conflict?
```
mmorhun, line 333: Not retrying will keep the finalyzer, won't it?

mmorhun (approved):
I think that the scaffold is ready, we might revise some retrying logic when nudge implementation will be in place.

stuartwdouglas:
I am moving this back to draft for now as I am in the process of actually making the nudges work, I don't think there is much point merging this without it being actually able to do something.

stuartwdouglas:
So I have the renovate job generated and testing out the config locally it seems to work. Ideally we would have e2e tests for this but it may be a bit involved, as it will require creating additional repos in the redhat-appstudio-qe org, and then setting up test branches that 'nudge' each other.

chmeliik:
Could you add some links to the relevant docs (if they're available publicly) to the first commit message? Could help give more context to the reviewers

stuartwdouglas:
There are lots of documents referenced from the JIRA ticket and its parent, I don't think we want to be adding google doc links to commit messages.

chmeliik:
> There are lots of documents referenced from the JIRA ticket and its parent, I don't think we want to be adding google doc links to commit messages.

Maybe https://github.com/redhat-appstudio/architecture/blob/main/ADR/0029-component-dependencies.md, if that's the canonical one?

stuartwdouglas:
It looks like the latest envtest failure might be related, investigating.

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,484 @@
...
+		log.Error(fmt.Errorf("unable to find %s param on PipelineRun, not performing nudge", ImageDigestParamName), "no image digest result")
+		return r.removePipelineFinalizer(ctx, pipelineRun)
+	}
+	parts := strings.Split(image, ":")
+	if len(parts) != 2 {
+		log.Error(fmt.Errorf("invalid image name %s", image), "invalid image format")
+		return r.removePipelineFinalizer(ctx, pipelineRun)
+	}
+	tag = parts[1]
+	repo := parts[0]
```
chmeliik, line 231: 1. Does IMAGE_URL always include a tag?
2. Port numbers (e.g. `image-registry.openshift-image-registry.svc:5000`) could break this too, but that may be an acceptable bug for now

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,484 @@
...
+	for i := range components.Items {
+		comp := components.Items[i]
+		if comp.Spec.Application == updatedComponent.Spec.Application && slices.Contains(updatedComponent.Spec.BuildNudgesRef, comp.Name) {
+			toUpdate = append(toUpdate, comp)
+		}
+	}
+	var nudgeErr error
+	immediateRetry, nudgeErr = r.UpdateFunction(ctx, r.Client, r.Scheme, r.EventRecorder, toUpdate, &BuildResult{BuiltImageRepository: repo, BuiltImageTag: tag, Digest: digest, Component: updatedComponent})
+	if nudgeErr != nil {
+		log.Error(nudgeErr, fmt.Sprintf("component update of as a result of a build of %s failed", updatedComponent.Name), l.Audit, "true")
```
chmeliik, line 254: `"component update of <something missing here?> as a result of a build of %s failed"`

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,484 @@
...
+
+		pipelineRun.Annotations[FailureCountAnnotationName] = strconv.Itoa(failureCount)
+
+		r.EventRecorder.Event(updatedComponent, corev1.EventTypeWarning, ComponentNudgeFailedEventType, fmt.Sprintf("component update failed as a result of a build for %s, retry %d/%d", updatedComponent.Name, failureCount, MaxAttempts))
+
+		if failureCount >= MaxAttempts {
+			// We are at the failure limit, nothing much we can do
+			log.Info("not retrying as max failure limit has been reached", l.Audit, "true")
+			return r.removePipelineFinalizer(ctx, pipelineRun)
+		}
```
chmeliik, line 278: Should we maybe do this check first, before trying to do the update? Isn't it possible that the reconcile will get requeued for a PipelineRun which has already reached max retries?

chmeliik (commented):
Mostly LGTM :+1: 

One general question: the [ADR](https://github.com/redhat-appstudio/architecture/blob/main/ADR/0029-component-dependencies.md) talks about nudging on PRs as well, here we only nudge on push. Has it changed since then, or is nudging on PRs to be implemented later?

stuartwdouglas:
@chmeliik we decided nudging on PR was out of scope for now, and i want to look at ways of doing testing that does not involve a throwing around heaps of unmergeable PRs. For now we are just going basic nudges.

stuartwdouglas:
I have no idea why the e2e test is failing here but passing in the e2e repo. The renovate job is saying that the base branch does not exist, even though you can see closed PRs into the base branch created by these tests, so it definitely did exist.

controllers/renovate_util.go
```diff
@@ -0,0 +1,324 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"os"
+	"regexp"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	log2 "sigs.k8s.io/controller-runtime/pkg/log"
```
mmorhun, line 22: Can we have better than `log2` name?

pkg/git/github/github_app.go
```diff
@@ -34,7 +34,8 @@ var NewGithubClientByApp func(appId int64, privateKeyPem []byte, repoUrl string)
 var NewGithubClientForSimpleBuildByApp func(appId int64, privateKeyPem []byte) (*GithubClient, error) = newGithubClientForSimpleBuildByApp
 
 var IsAppInstalledIntoRepository func(ghclient *GithubClient, repoUrl string) (bool, error) = isAppInstalledIntoRepository
-var GetAppInstallations func(githubAppIdStr string, appPrivateKeyPem []byte) ([]ApplicationInstallation, string, error) = getAppInstallations
+var GetAllAppInstallations func(githubAppIdStr string, appPrivateKeyPem []byte) ([]ApplicationInstallation, string, error) = getAppInstallations
+var GetAppInstallationsForRepository func(githubAppIdStr string, appPrivateKeyPem []byte, owner string, repo string) (*ApplicationInstallation, string, error) = getAppInstallationsForRepository
```
mmorhun, line 38: For public functions the preferred way is to pass `repoUrl` and retrieve owner/repo inside. That would allow generic interface with other git providers.

pkg/git/github/github_app.go
```diff
@@ -270,6 +271,57 @@ func getAppInstallations(githubAppIdStr string, appPrivateKeyPem []byte) ([]Appl
...
+	if err != nil {
+		// Inability to create transport based on a private key indicates that the key is bad formatted
+		return nil, "", boerrors.NewBuildOpError(boerrors.EGitHubAppMalformedPrivateKey, err)
+	}
+	client := github.NewClient(&http.Client{Transport: itr})
+	githubApp, _, err := client.Apps.Get(context.Background(), "")
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to load GitHub app metadata, %w", err)
+	}
+	slug := (githubApp.GetSlug())
```
mmorhun, line 291: Extra brackets

chmeliik:
I had some comments here before they got lost in the github history, not sure if you've had a chance to address them: https://github.com/redhat-appstudio/build-service/pull/210#discussion_r1415559005 (and the following ones)

controllers/renovate_util.go
```diff
@@ -0,0 +1,324 @@
...
+	"strings"
+	"time"
+)
+
+// GetAllGithubInstallations gets installations by iterating over every installation, should be used for cases where we expect to be touching most installed repos
+func GetAllGithubInstallations(ctx context.Context, client client.Client, eventRecorder record.EventRecorder, componentList []v1alpha1.Component) (string, []installationStruct, error) {
+	log := log2.FromContext(ctx)
+	// Check if GitHub Application is used, if not then skip
+	pacSecret := v1.Secret{}
+	globalPaCSecretKey := types.NamespacedName{Namespace: buildServiceNamespaceName, Name: prepare.PipelinesAsCodeSecretName}
```
chmeliik, line 32: General review-ability observation: next time, it would be nice to split the PR into commits that make it easier to review. Specifically, some of the changes in this PR just move code from one place to another - it would have been nice to do that in a separate commit and explain that there's no change in functionality, that it's just moving code around. (I'm not actually sure that there are no changes in the code that got moved, just assuming that's true).

Having it all in a single commit makes it hard to focus on the things that matter, first one has to figure out what is or isn't a relevant change

controllers/pac_pipelinerun_pruner_controller.go
```diff
@@ -65,7 +65,7 @@ func (r *PaCPipelineRunPrunerReconciler) SetupWithManager(mgr ctrl.Manager) erro
 		Complete(r)
 }
 
-//+kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns,verbs=get;list;watch;delete;deletecollection
+//+kubebuilder:rbac:groups=tekton.dev,resources=pipelineruns,verbs=get;list;watch;delete;deletecollection;update;patch
```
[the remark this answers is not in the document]
stuartwdouglas: Apparently not, I tried adding them to the new controller and the ones here took precedence. Seemed to be an issue with the controller gen.

controllers/component_dependency_update_controller.go
stuartwdouglas: I will change it to split manually lastIndexOf ':'. There should never be a digest as this is a separate field. In practice this is always redhat-user-workloads anyway.

controllers/component_dependency_update_controller.go
stuartwdouglas: No, because the code that would update the failure count will also add the completed annotation. If you see a PipelineRun that has hit the failure count it will already be marked as done and ignored by the reconcile loop.

chmeliik: Right, but `removePipelineFinalizer` can fail as well. And it might not even be that uncommon due to the number of controllers that reconcile on and update PipelineRuns?

So IIUC the operation that follows immediately after "not retrying as max failure limit has been reached" can fail and cause a retry

chmeliik: Maybe putting the max retries check before the `r.UpdateFunction()` call could avoid retrying more than intended (unless the operation that sets the failure count annotation fails)

controllers/component_dependency_update_controller.go
chmeliik: bump

Skipping already-processed pipeline makes sense to me as well

controllers/component_dependency_update_controller.go
```diff
@@ -0,0 +1,511 @@
...
+		return r.removePipelineFinalizer(ctx, pipelineRun)
+	}
+	if digest == "" {
+		log.Error(fmt.Errorf("unable to find %s param on PipelineRun, not performing nudge", ImageDigestParamName), "no image digest result")
+		return r.removePipelineFinalizer(ctx, pipelineRun)
+	}
+	tag := ""
+	repo := image
+	index := strings.LastIndex(image, ":")
+	if index != -1 {
```
mmorhun, line 239: To not to have func wide variable `index`, you can do:
```
if index := strings.LastIndex(image, ":"); index != -1 {
```

mmorhun (approved):

controllers/component_dependency_update_controller.go
stuartwdouglas: There is nothing we can do about this, the removal of the finalizer and setting of the update count is the same update operation. If it does fail and a retry happens more than expected that is not a huge deal, especially given that failure will be very rare in normal circumstances. Failure in this case is failure to launch the renovate job, we don't reconcile on jobs and attempt to check the result.

PR #211, opened 2023-11-09
Use secret from component instead of SPITokenBinding to access private git repository

mmorhun:
Changes the way Build Service expects git credentials for simple builds: instead of searching for SPIAccessTokenBindings (deprecated) use the secret specified in the Component spec.


controllers/component_build_controller_simple_build.go
```diff
@@ -136,14 +135,18 @@ func (r *ComponentBuildReconciler) getBuildGitInfo(ctx context.Context, componen
...
-		if err != nil {
-			log.Error(err, "failed to find git secret name")
-			return nil, err
-		}
+		gitSecretName = component.Spec.Secret
 		if gitSecretName == "" {
-			// The repository is private and no git secret provided
-			return nil, boerrors.NewBuildOpError(boerrors.EComponentGitSecretMissing, nil)
+			// The repository is private and no git secret specified
+			return nil, boerrors.NewBuildOpError(boerrors.EComponentGitSecretNotSpecified, nil)
```
mkosiarc, line 141: What would be the difference between `boerrors.EComponentGitSecretMissing`  and `boerrors.EComponentGitSecretNotSpecified` here?

mmorhun: `NotSpecified` happens when the secret name was not given, where as `Missing` when secret name is specified, but the secret itself does not exist.

mkosiarc: I see the description in `pkg/boerrors/perror.go` but I am not sure I completely understand

mkosiarc:
LGTM except for the failing CI, is it related to this PR?

controllers/component_build_controller_simple_build.go
tkdchen: The diagram shows the process is:

UI HAC creates RemoteSecret -> RemoteSecret creates the Secret -> UI HAC update the Secret name to component

Should we treat the this error as transient error and emit Event as well?

tkdchen, line 141: The diagram shows the process is:

UI HAC creates RemoteSecret -> RemoteSecret creates the Secret -> UI HAC update the Secret name to component

Should we treat the this error as transient error and emit Event as well?

mmorhun: it might be both, but I'd stick to permanent, because user can edit Component spec manually.

mkosiarc (approved):

PR #212, opened 2023-11-15
Update dependencies

mmorhun:
This PR updates:
- images used by the operator
- operator dependencies


controllers/component_build_controller.go
```diff
@@ -217,6 +217,9 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, nil
 	}
 
+	log = log.WithValues("ComponentGitSource", component.Spec.Source.GitSource)
+	ctx = ctrllog.IntoContext(ctx, log)
+
```
tkdchen, line 222: Is this necessary for the update?

Dockerfile
```diff
@@ -18,9 +20,16 @@ COPY pkg/ pkg/
...
+# https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5
+FROM registry.access.redhat.com/ubi9/ubi-minimal:9.3
 COPY --from=builder /opt/app-root/src/manager /
 USER 65532:65532
 
+LABEL description="RHTAP Build Service operator"
+LABEL io.k8s.description="RHTAP Build Service operator"
+LABEL io.k8s.display-name="build-service-operator"
+LABEL io.openshift.tags="rhtap"
+LABEL summary="RHTAP Build Service"
```
chmeliik, line 33: Could the labels be in a separate commit?

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -15,8 +15,9 @@ spec:
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.13.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.15.0
         args:
+        - "--http2-disable"
```
chmeliik, line 20: I guess it's still not safe to explain why we're adding `--http2-disable`? :disappointed:

main.go
```diff
@@ -199,22 +202,20 @@ func getCacheExcludedObjectsTypes() []client.Object {
 	}
 }
 
-func getCacheFuncOptions() (*cache.Options, error) {
+func getCacheOptions() cache.Options {
 	componentPipelineRunRequirement, err := labels.NewRequirement(controllers.ComponentNameLabelName, selection.Exists, []string{})
 	if err != nil {
-		return nil, err
+		panic(err)
```
chmeliik, line 208: Why switch from returning an error to `panic`?

mmorhun: Because the operation above is safe. It can error if we configure it wrongly, but that should be fixed in the same commit and shouldn't bother the caller.

controllers/component_build_controller.go
mmorhun: Not really. It's just a minor improvement for logging.

main.go
chmeliik: Ack :+1: 

Could you add a comment about that or explain it in the commit message?

controllers/component_build_controller.go
mkosiarc: I would also include it in separate commit in that case

mmorhun: ok

Dockerfile
mmorhun: ok

main.go
mmorhun: ok

chmeliik (approved):
LGTM :+1: 

CI needs a rerun

main.go
```diff
@@ -109,21 +110,23 @@ func main() {
...
-	if err != nil {
-		setupLog.Error(err, "unable to create cache options")
-		os.Exit(1)
+	clientOpts := client.Options{
+		Cache: &client.CacheOptions{
+			DisableFor: getCacheExcludedObjectsTypes(),
+		},
+	}
+	metricsOpts := server.Options{
+		BindAddress: metricsAddr,
```
mkosiarc, line 119: How is this code change related to the updated operator dependencies?

mmorhun: Breaking changes in the configuration structures.

mkosiarc (approved):

PR #214, opened 2023-11-15
Drop v1beta1 selector support

chmeliik:
Don't support the name+bundle style, support only the resolver+params
style for BuildPipelineSelector pipelineRefs.


mmorhun (approved):

chmeliik:
Converting to draft until https://github.com/redhat-appstudio/build-definitions/pull/677 is merged.

I think that's the last place where RHTAP uses the old selector style :crossed_fingers:

PR #215, opened 2023-11-21
when removing PaC secret, ensure that repository still exists, so we

rcerven:
actually know secret name

[STONEBLD-1968](https://issues.redhat.com//browse/STONEBLD-1968)

### DIFF
--- a/controllers/component_build_controller_pac.go
+++ b/controllers/component_build_controller_pac.go
@@ -340,46 +340,49 @@ func (r *ComponentBuildReconciler) cleanupPaCRepositoryIncomingsAndSecret(ctx co
 	if err != nil {
 		return err
 	}
-	incomingSecretName := ""
-	if repository != nil {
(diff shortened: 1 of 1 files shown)


mmorhun (approved):

controllers/component_build_controller_pac.go
```diff
@@ -340,9 +340,8 @@ func (r *ComponentBuildReconciler) cleanupPaCRepositoryIncomingsAndSecret(ctx co
 	if err != nil {
 		return err
 	}
-	incomingSecretName := ""
 	if repository != nil {
-		incomingSecretName = fmt.Sprintf("%s%s", repository.Name, pacIncomingSecretNameSuffix)
+		incomingSecretName := fmt.Sprintf("%s%s", repository.Name, pacIncomingSecretNameSuffix)
```
chmeliik, line 344: Instead of nesting the rest of this function under the if, could we `return` early when `repository == nil`?

rcerven: sure

chmeliik (approved):
LGTM :+1:

PR #222, opened 2023-12-07
Disable semantic commit message for renovate

mmorhun:
This PR disables semantic commit message style fro renovate bot commits.
This is done to prevent renovate updating own PR in a loop when it fails to properly auto-detect whether to use semantic commit messages.

### DIFF
--- a/controllers/git_tekton_resources_renovater.go
+++ b/controllers/git_tekton_resources_renovater.go
@@ -240,6 +240,7 @@ func generateConfigJS(slug string, repositories []renovateRepository) string {
 				branchName: "rhtap/references/{{baseBranch}}",
 				commitMessageExtra: "",
 				commitMessageTopic: "RHTAP references",
+				semanticCommits: "enabled",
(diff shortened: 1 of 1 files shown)


rcerven (approved):

chmeliik:
By default, renovate tries to detect which style to use based on the commit history of the repo?

I'm wondering if this change will make our renovate PRs fail CI checks for repos that use semantic commits

tkdchen (approved):

mmorhun:
> By default, renovate tries to detect which style to use based on the commit history of the repo?

Yes, but sometimes it fails and does strange things:
image

> I'm wondering if this change will make our renovate PRs fail CI checks for repos that use semantic commits

@chmeliik  it might. The idea is to not let renovate autodetect, so the situation on the screenshot above won't happen. We may turn semantic commits style on, so it won't fail the checks on repositories that require them.

chmeliik (approved):
LGTM :+1:

tkdchen (approved):

PR #223, opened 2023-12-07
reporting request times for each action to prometheus for SLO/SLI

rcerven:
[STONEBLD-1773](https://issues.redhat.com//browse/STONEBLD-1773)
[STONEBLD-1774](https://issues.redhat.com//browse/STONEBLD-1774)

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -75,33 +75,76 @@ const (
 )
 
 var (
-	simpleBuildPipelineCreationTimeMetric       prometheus.Histogram
-	pipelinesAsCodeComponentProvisionTimeMetric prometheus.Histogram
+	componentOnboardingTimeMetric                 prometheus.Histogram
+	simpleBuildPipelineCreationTimeMetric         prometheus.Histogram
+	pipelinesAsCodeComponentProvisionTimeMetric   prometheus.Histogram
+	pipelinesAsCodeComponentUnconfigureTimeMetric prometheus.Histogram
+	pushPipelineRebuildTriggerTimeMetric          prometheus.Histogram
+	componentTimesForMetrics                      = map[string]componentMetricsInfo{}
 )
+
+type componentMetricsInfo struct {
+	startTimestamp  time.Time
+	requestedAction string
+}
 
 func initMetrics() error {
 	buckets := getProvisionTimeMetricsBuckets()
 
+	componentOnboardingTimeMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: metricsNamespace,
+		Subsystem: metricsSubsystem,
+		Buckets:   buckets,
+		Name:      "component_onboarding_time",
+		Help:      "The time in seconds spent from the moment of Component creation till simple build pipeline submission, or PaC provision.",
+	})
 	simpleBuildPipelineCreationTimeMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsSubsystem,
 		Buckets:   buckets,
-		Name:      "initial_build_pipeline_creation_time",
-		Help:      "The time in seconds spent from the moment of Component creation till the initial build pipeline submission.",
+		Name:      "simple_build_pipeline_creation_time",
+		Help:      "The time in seconds spent from the moment of requesting simple build for Component till build pipeline submission.",
 	})
 	pipelinesAsCodeComponentProvisionTimeMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
 		Namespace: metricsNamespace,
 		Subsystem: metricsSubsystem,
 		Buckets:   buckets,
 		Name:      "PaC_configuration_time",
-		Help:      "The time in seconds spent from the moment of Component creation till Pipelines-as-Code configuration done in the Component source repository.",
+		Help:      "The time in seconds spent from the moment of requesting PaC provision till Pipelines-as-Code configuration done in the Component source repository.",
+	})
+	pipelinesAsCodeComponentUnconfigureTimeMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: metricsNamespace,
+		Subsystem: metricsSubsystem,
+		Buckets:   buckets,
+		Name:      "PaC_unconfiguration_time",
+		Help:      "The time in seconds spent from the moment of requesting PaC unprovision till Pipelines-as-Code configuration is removed in the Component source repository.",
+	})
+	pushPipelineRebuildTriggerTimeMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: metricsNamespace,
+		Subsystem: metricsSubsystem,
+		Buckets:   buckets,
+		Name:      "Push_pipeline_rebuild_trigger_time",
+		Help:      "The time in seconds spent from the moment of requesting push pipeline rebuild till Pipelines-as-Code API trigger.",
 	})
 
-	if err := metrics.Registry.Register(simpleBuildPipelineCreationTimeMetric); err != nil {
-		return fmt.Errorf("failed to register the initial_build_pipeline_creation_time metric: %w", err)
+	if err := metrics.Registry.Register(componentOnboardingTimeMetric); err != nil {
+		return fmt.Errorf("failed to register the initial_build_or_pac_provision_creation_time metric: %w", err)
 	}
+
+	if err := metrics.Registry.Register(simpleBuildPipelineCreationTimeMetric); err != nil {
+		return fmt.Errorf("failed to register the simple_build_pipeline_creation_time metric: %w", err)
+	}
+
 	if err := metrics.Registry.Register(pipelinesAsCodeComponentProvisionTimeMetric); err != nil {
 		return fmt.Errorf("failed to register the PaC_configuration_time metric: %w", err)
+	}
+
+	if err := metrics.Registry.Register(pipelinesAsCodeComponentUnconfigureTimeMetric); err != nil {
+		return fmt.Errorf("failed to register the PaC_unconfiguration_time metric: %w", err)
+	}
+
+	if err := metrics.Registry.Register(pushPipelineRebuildTriggerTimeMetric); err != nil {
+		return fmt.Errorf("failed to register the Push_pipeline_rebuild_trigger_time metric: %w", err)
 	}
 
 	return nil
@@ -176,6 +219,20 @@ func (r *ComponentBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+func updateMetricsTimes(componentIdForMetrics string, requestedAction string, reconcileStartTime time.Time) {
+	componentInfo, timeRecorded := componentTimesForMetrics[componentIdForMetrics]
+
+	// first reconcile
+	if !timeRecorded {
+		componentTimesForMetrics[componentIdForMetrics] = componentMetricsInfo{startTimestamp: reconcileStartTime, requestedAction: requestedAction}
+	} else {
+		// new different request
+		if componentInfo.requestedAction != requestedAction {
+			componentTimesForMetrics[componentIdForMetrics] = componentMetricsInfo{startTimestamp: reconcileStartTime, requestedAction: requestedAction}
+		}
+	}
+}
+
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/status,verbs=get;list;watch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=buildpipelineselectors,verbs=get;list;watch;create;update;patch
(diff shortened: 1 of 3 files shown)


chmeliik (commented):
Could you explain (preferably in the commit message) the logic behind how this works? E.g. how it handles counting the time across multiple Reconcile retries, how it distinguishes between a retry vs. a new request etc.

I'm having trouble convincing myself that the implementation is correct

controllers/component_build_controller.go
```diff
@@ -342,6 +405,11 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			}
 		} else {
 			simpleBuildStatus.BuildStartTime = time.Now().Format(time.RFC1123)
+			simpleBuildPipelineCreationTimeMetric.Observe(time.Since(componentTimesForMetrics[componentIdForMetrics].startTimestamp).Seconds())
+			// initial build upon component creation
+			if initialBuildStatus.PaC == nil && initialBuildStatus.Simple == nil {
+				initialComponentCreationBuildOrPaCProvisionTimeMetric.Observe(time.Since(componentTimesForMetrics[componentIdForMetrics].startTimestamp).Seconds())
+			}
```
chmeliik, line 412: What is the difference between these two metrics, other than one of them mixing both simple builds and PaC provisions? Specifically, what is the purpose of the check `if initialBuildStatus.PaC == nil && initialBuildStatus.Simple == nil`?

controllers/component_build_controller.go
```diff
@@ -487,6 +578,8 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 	log.Info(fmt.Sprintf("updated component after build request: %s", requestedAction), l.Action, l.ActionUpdate)
+	// remove component from metrics map
+	delete(componentTimesForMetrics, componentIdForMetrics)
```
chmeliik, line 582: Is there a specific reason to delete the component from the global map _after_ doing the `Update` above, where we remove the request annotation? If we made it all the way here in the code, then the simple build / the PaC provision / etc. already succeeded no?

rcerven: if component isn't updated with removed request annotation 

and if update fails it will trigger another reconcile which will process request again

controllers/component_build_controller.go
chmeliik: Ah, so `initialBuildStatus.PaC == nil && initialBuildStatus.Simple == nil` means that the component hasn't had any initial builds or PaC provisions yet?

That makes sense. Could you add a comment to explain that?

controllers/component_build_controller.go
chmeliik: Ah, that makes sense.

I guess the question is what exactly we want the SLIs to represent. The way they were worded originally, the time until build submission / PR submission would be what we want to measure (and also what the user probably cares about).

> Time taken from simple build request till the build pipeline is submitted.

> Time taken from PaC provision request till Component is provisioned for PaC builds. The last step (trigger) is that PR is created in GitHub.

But counting the time until the annotation is removed makes sense to me as well, since the controller's work technically isn't done until then. @mmorhun any thoughts?

controllers/component_build_controller.go
```diff
@@ -75,35 +75,78 @@ const (
...
 func initMetrics() error {
 	buckets := getProvisionTimeMetricsBuckets()
 
+	initialComponentCreationBuildOrPaCProvisionTimeMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: metricsNamespace,
+		Subsystem: metricsSubsystem,
+		Buckets:   buckets,
+		Name:      "initial_build_or_pac_provision_creation_time",
+		Help:      "The time in seconds spent from the moment of Component creation till simple build pipeline submission, or PaC provision.",
+	})
```
chmeliik, line 100: Would it make sense to split this one into two metrics, one for simple builds and one for PaC provisions? Otherwise we won't have the data about which one is slower

controllers/component_build_controller.go
rcerven: updated

controllers/component_build_controller.go
rcerven: keep in mind that time difference between "action done" and "removing annotation" is minimal

also IF we would count success before removing annotation we would could be counting (build which failed on removal of annotation) multiple times

chmeliik: > also IF we would count success before removing annotation we would could be counting (build which failed on removal of annotation) multiple times

Good point :+1:

controllers/component_build_controller.go
chmeliik: hmm, ok

chmeliik (approved):
LGTM :+1:

controllers/component_build_controller.go
```diff
@@ -329,6 +391,9 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	switch requestedAction {
 	case BuildRequestTriggerSimpleBuildAnnotationValue:
+		updateMetricsTimes(componentIdForMetrics, requestedAction, reconcileStartTime)
+		initialBuildStatus := readBuildStatus((&component))
```
tkdchen, line 395: Why pass the component like this `(&component)`? The parentheses should not be required.

controllers/component_build_controller.go
tkdchen: `simpleBuildPipelineCreationTimeMetric` could be part of this metric. Creating a component is actually one way to trigger/request a simple build or PaC provision.

My understanding the consideration of this metric is it is just for the time counting for the first request upon Component creation. Whereas, both `simpleBuildPipelineCreationTimeMetric` and `pipelinesAsCodeComponentProvisionTimeMetric` collect the time for the first request (by creation) and subsequent requests.

However, if we need `simpleBuildPipelineCreationTimeMetric`, it would be good to split it even we have the other two. The overall time of PaC provision could be affected by more aspects (including the external APIs) than the simple build.

controllers/component_build_controller.go
```diff
@@ -75,35 +75,78 @@ const (
...
 
 var (
-	simpleBuildPipelineCreationTimeMetric       prometheus.Histogram
-	pipelinesAsCodeComponentProvisionTimeMetric prometheus.Histogram
+	initialComponentCreationBuildOrPaCProvisionTimeMetric prometheus.Histogram
+	simpleBuildPipelineCreationTimeMetric                 prometheus.Histogram
+	pipelinesAsCodeComponentProvisionTimeMetric           prometheus.Histogram
+	pipelinesAsCodeComponentUnconfigureTimeMetric         prometheus.Histogram
+	pipelinesAsCodeComponentTriggerTimeMetric             prometheus.Histogram
+	componentTimesForMetrics                                  = map[string]componentMetricsInfo{}
```
tkdchen, line 83: Any reason to use this mapping? In what cases multiple components are recorded?

controllers/component_build_controller.go
```diff
@@ -342,6 +407,11 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			}
 		} else {
 			simpleBuildStatus.BuildStartTime = time.Now().Format(time.RFC1123)
+			simpleBuildPipelineCreationTimeMetric.Observe(time.Since(componentTimesForMetrics[componentIdForMetrics].startTimestamp).Seconds())
+			// initial build upon component creation (doesn't have either build status)
+			if initialBuildStatus.PaC == nil && initialBuildStatus.Simple == nil {
```
tkdchen, line 412: Suggest not use words "initial build" here to avoid possible confusion, as build-service has a concept of initial build. `buildStatus` should be enough to describe what the variable references.

controllers/component_build_controller.go
```diff
@@ -75,35 +75,78 @@ const (
...
+	simpleBuildPipelineCreationTimeMetric                 prometheus.Histogram
+	pipelinesAsCodeComponentProvisionTimeMetric           prometheus.Histogram
+	pipelinesAsCodeComponentUnconfigureTimeMetric         prometheus.Histogram
+	pipelinesAsCodeComponentTriggerTimeMetric             prometheus.Histogram
+	componentTimesForMetrics                                  = map[string]componentMetricsInfo{}
 )
 
+type componentMetricsInfo struct {
+	startTimestamp time.Time
+	requestType    string
```
tkdchen, line 88: By reading through the code, I know that `requestType` is the `requestAction`. How about use same name or give it a short description?

controllers/component_build_controller.go
```diff
@@ -380,9 +452,13 @@ func (r *ComponentBuildReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 			if reconcileRequired {
 				return ctrl.Result{Requeue: true}, nil
 			}
+			pipelinesAsCodeComponentTriggerTimeMetric.Observe(time.Since(componentTimesForMetrics[componentIdForMetrics].startTimestamp).Seconds())
 		}
 
 	case BuildRequestConfigurePaCAnnotationValue:
+		updateMetricsTimes(componentIdForMetrics, requestedAction, reconcileStartTime)
+		initialBuildStatus := readBuildStatus((&component))
```
mmorhun, line 460: The same as above

controllers/component_build_controller.go
rcerven: removed

controllers/component_build_controller.go
rcerven: we have to record all components, because build can have transient error on which it will run another reconcile, so we have to count that as 1 run

controllers/component_build_controller.go
rcerven: I'd say 'initial build upon component creation' isn't confusing, also I wanted to make it clear for everybody

Adam wanted more clear explanation and you want less

what I have is good I'd say

controllers/component_build_controller.go
rcerven: sure

controllers/component_build_controller.go
```diff
@@ -75,35 +75,78 @@ const (
...
+
 func initMetrics() error {
 	buckets := getProvisionTimeMetricsBuckets()
 
+	initialComponentCreationBuildOrPaCProvisionTimeMetric = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Namespace: metricsNamespace,
+		Subsystem: metricsSubsystem,
+		Buckets:   buckets,
+		Name:      "initial_build_or_pac_provision_creation_time",
+		Help:      "The time in seconds spent from the moment of Component creation till simple build pipeline submission, or PaC provision.",
```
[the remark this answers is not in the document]
rcerven: it is needed, when those closing braces are on the next line

and it looks better with closing ones on new line

controllers/component_build_controller.go
```diff
@@ -176,6 +219,20 @@ func (r *ComponentBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
+func updateMetricsTimes(componentIdForMetrics string, requestedAction string, reconcileStartTime time.Time) {
+	componentInfo, timeRecorded := componentTimesForMetrics[componentIdForMetrics]
+
+	// first reconcile
+	if !timeRecorded {
+		componentTimesForMetrics[componentIdForMetrics] = componentMetricsInfo{startTimestamp: reconcileStartTime, requestType: requestedAction}
+	} else {
```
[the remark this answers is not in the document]
rcerven: it doesn't simplify it by much but return can make it confusing

I'd just keep it so it is more clear

controllers/component_build_controller.go
```diff
@@ -75,35 +75,78 @@ const (
 )
 
 var (
-	simpleBuildPipelineCreationTimeMetric       prometheus.Histogram
-	pipelinesAsCodeComponentProvisionTimeMetric prometheus.Histogram
+	initialComponentCreationBuildOrPaCProvisionTimeMetric prometheus.Histogram
```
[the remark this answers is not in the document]
rcerven: renamed

controllers/component_build_controller.go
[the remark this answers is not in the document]
rcerven: I mean, it is used for metrics, but it is really just buildstatus from component, which makes it usable later for other things

controllers/component_build_controller.go
rcerven: same as above :)

controllers/component_build_controller.go
chmeliik: So @tkdchen you're also +1 to have two separate onboarding metrics?

So that we would have
* simple component onboarding (component creation with a simple build request)
* PaC component onboarding (component creation with a PaC provision request)
* simple pipeline creation (any request for a simple build)
* PaC provisioning (any PaC provisioning request)

controllers/component_build_controller.go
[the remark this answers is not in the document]
mmorhun: I don't think it's used, but it confuse, IMO.

controllers/component_build_controller.go
chmeliik: Ack, that's fair.

Maybe the general metrics for simple builds and PaC provisions should exclude the onboarding cases then? Otherwise they could skew the metrics a bit by including these longer times

mmorhun (approved):

controllers/component_build_controller.go
rcerven: I don't think that the onboarding cases will took a lot more, but
excluded them from those general metrics and counting it only for onboarding one

PR #224, opened 2023-12-07
Add script for setting appstudio_workspace params

chmeliik:
[STONEBLD-2008](https://issues.redhat.com//browse/STONEBLD-2008)

For newly created Repository objects, build-service will be setting the appstudio_workspace param. For Repository objects created before this feature was deployed, the appstudio_workspace param will be missing.

Add a script that automatically sets the appstudio_workspace param for all Repositories that don't have it set correctly yet.

The script requires python>=3.8 and depends on kubectl or oc (and nothing else).

### DIFF
--- a/hack/add_appstudio_workspace_param.py
+++ b/hack/add_appstudio_workspace_param.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python
+
+"""
+Add the appstudio_workspace parameter to Pipelines as Code repository objects.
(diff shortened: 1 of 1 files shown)


chmeliik:
@tkdchen @mmorhun do you think this is the right place for the script? Or should I put it in some SRE repo?

chmeliik:
Per offline discussion, the script will be a one-time action so we probably don't need it in any repo at all. Will publish as a Github gist and link to STONEBLD-2008

PR #225, opened 2023-12-08
fix failing renovater test

rcerven:
test which doesn't trigger any jobs should run at the end, because when it is first it can race with 2nd test and while 1st renovate reconcile is running, it can already renovate for component from 2nd test

### DIFF
--- a/controllers/git_tekton_resources_renovater_test.go
+++ b/controllers/git_tekton_resources_renovater_test.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"os"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -50,21 +51,6 @@ var _ = Describe("Git tekton resources renovater", func() {
 			deleteJobs(buildServiceNamespaceName)
 			os.Unsetenv(InstallationsPerJobEnvName)
 			deleteSecret(pacSecretKey)
-		})
-
-		It("It should not trigger job", func() {
-			installedRepositoryUrls := []string{
-				"https://github/test/repo1",
-				"https://github/test/repo2",
-			}
-			github.GetAppInstallations = func(appIdStr string, privateKeyPem []byte) ([]github.ApplicationInstallation, string, error) {
-				repositories := generateRepositories(installedRepositoryUrls)
-				return []github.ApplicationInstallation{generateInstallation(repositories)}, "slug", nil
-			}
-			componentNamespacedName := createComponentForPaCBuild(getComponentData(componentConfig{componentKey: types.NamespacedName{Name: "testnottrigger"}, gitURL: "https://github/test/repo3"}))
-			createDefaultBuildPipelineRunSelector(defaultSelectorKey)
-			Eventually(listJobs).WithArguments(buildServiceNamespaceName).WithTimeout(timeout).Should(BeEmpty())
-			deleteComponent(componentNamespacedName)
 		})
 
 		It("It should trigger job", func() {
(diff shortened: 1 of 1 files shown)


controllers/git_tekton_resources_renovater_test.go
```diff
@@ -166,5 +152,22 @@ var _ = Describe("Git tekton resources renovater", func() {
...
+				"https://github/test/repo2",
+			}
+			github.GetAppInstallations = func(appIdStr string, privateKeyPem []byte) ([]github.ApplicationInstallation, string, error) {
+				repositories := generateRepositories(installedRepositoryUrls)
+				return []github.ApplicationInstallation{generateInstallation(repositories)}, "slug", nil
+			}
+			componentNamespacedName := createComponentForPaCBuild(getComponentData(componentConfig{componentKey: types.NamespacedName{Name: "testnottrigger"}, gitURL: "https://github/test/repo3"}))
+			createDefaultBuildPipelineRunSelector(defaultSelectorKey)
+			// try to wait for renovate Reconcile at least start
+			time.Sleep(2 * time.Second)
```
tkdchen, line 168: What does it mean by "at least start"?

For what to wait for renovate Reconcile as least start?

tkdchen (commented):
Generally, we should ensure the test specs `It` should not depend on each other. What do you think to stop the Reconcile rather than rescheduling the Reconcile for integration tests? We actually don't test the reschedule behavior in integration tests and I don't think we have to.

controllers/git_tekton_resources_renovater_test.go
rcerven: well JUST in case it would really create some jobs ;D

controllers/git_tekton_resources_renovater_test.go
```diff
@@ -166,5 +152,22 @@ var _ = Describe("Git tekton resources renovater", func() {
...
+			Eventually(listJobs).WithArguments(buildServiceNamespaceName).WithTimeout(timeout).Should(BeEmpty())
+			deleteComponent(componentNamespacedName)
```
mmorhun, line 170: Better to move it into AfterEach

controllers/git_tekton_resources_renovater_test.go
mmorhun: I don't think it's needed.

chmeliik:
> test which doesn't trigger any jobs should run at the end, because when it is first it can race with 2nd test and while 1st renovate reconcile is running, it can already renovate for component from 2nd test

Info that I'm missing to understand this change:
* _How_ was the renovater test failing? Was it that there were 2 jobs and it expected only 1?
* What specifically causes ^
 
My guess:
* the 1st test does something (creates the selector?) that triggers the Renovate Reconcile
* the 1st test checks that there are no jobs, this check succeeds immediately
* the 2nd test starts, but the first Reconcile still hasn't finished
* the 2nd test creates a new Component before the first Reconcile has finished, the first Reconcile creates a job for this component
* the 2nd test creates another selector, this triggers another Reconcile
* the second Reconcile creates a second job

If the guess is correct, then the root of the problem is that the 1st test doesn't actually test anything? Since the "there are no jobs" check finishes before the Reconcile?

controllers/git_tekton_resources_renovater_test.go
rcerven: I mean without any wait for renovate, this test doesn't test anything

so with this test makes at least some sense, and it isn't such huge delay

rcerven:
> > test which doesn't trigger any jobs should run at the end, because when it is first it can race with 2nd test and while 1st renovate reconcile is running, it can already renovate for component from 2nd test
> 
> Info that I'm missing to understand this change:
> 
>     * _How_ was the renovater test failing? Was it that there were 2 jobs and it expected only 1?
> 
>     * What specifically causes ^
> 
> 
> My guess:
> 
>     * the 1st test does something (creates the selector?) that triggers the Renovate Reconcile
> 
>     * the 1st test checks that there are no jobs, this check succeeds immediately
> 
>     * the 2nd test starts, but the first Reconcile still hasn't finished
> 
>     * the 2nd test creates a new Component before the first Reconcile has finished, the first Reconcile creates a job for this component
> 
>     * the 2nd test creates another selector, this triggers another Reconcile
> 
>     * the second Reconcile creates a second job
> 
> 
> If the guess is correct, then the root of the problem is that the 1st test doesn't actually test anything? Since the "there are no jobs" check finishes before the Reconcile?

yes basically test doesn't trigger any renovate jobs (but triggers renovate reconcile), and sometimes 2nd test is too fast so it creates component and that is used already for that 1st renovator job

controllers/git_tekton_resources_renovater_test.go
rcerven: I can't do that as not all tests are using it

mmorhun (approved):

PR #226, opened 2023-12-11
Make cache update race condition handler optional

tkdchen:


mmorhun:
Does it solve issues with integration tests instabilities?

PR #228, opened 2023-12-12
fix failing test about linked secrets in repository

rcerven:
[STONEBLD-1998](https://issues.redhat.com//browse/STONEBLD-1998)

### DIFF
--- a/controllers/component_build_controller_test.go
+++ b/controllers/component_build_controller_test.go
@@ -685,6 +685,11 @@ var _ = Describe("Component initial build controller", func() {
 			component.Spec.ContainerImage = generatedImageRepo
 			Expect(k8sClient.Update(ctx, component)).To(Succeed())
 
+			// wait also for image finalizer on component
+			// image-registry-secret-sa-link.component.appstudio.openshift.io/finalizer
+			// otherwise when finalizer isn't there secret won't be unlinked from SA (when component is removed)
+			waitImageRegistrySecretLinkFinalizerOnComponent(resourcePacPrepKey)
+
 			Eventually(func() bool {
 				component = getComponent(resourcePacPrepKey)
 				return component.Spec.ContainerImage == generatedImageRepo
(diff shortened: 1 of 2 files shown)


controllers/component_build_controller_test.go
```diff
@@ -685,6 +685,11 @@ var _ = Describe("Component initial build controller", func() {
 			component.Spec.ContainerImage = generatedImageRepo
 			Expect(k8sClient.Update(ctx, component)).To(Succeed())
 
+			// wait also for image finalizer on component
+			// image-registry-secret-sa-link.component.appstudio.openshift.io/finalizer
+			// otherwise when finalizer isn't there secret won't be unlinked from SA (when component is removed)
+			waitImageRegistrySecretLinkFinalizerOnComponent(resourcePacPrepKey)
+
 			Eventually(func() bool {
```
tkdchen, line 693: So, the above update

```go
Expect(k8sClient.Update(ctx, component)).To(Succeed())
```

triggers a new reconcile to provision the PaC, which links the secret and removes `request` annotation eventually.

I think we can drop this `Eventually` call which checks the `ContainerImage`, because we just care about the finalizer. And if it is necessary to ensure the provision is done, it would be good to check the `request` annotation is removed or the `done` message in status, which are the last thing to be updated into a component during a reconcile.

tkdchen (commented):
LGTM

controllers/component_build_controller_test.go
rcerven: well PaC provision reconcile is already in progress, then we update annotation (so in the next reconcile finalizer will be added)

I don't want to remove that eventually since it was there 

we don't care tbh about done message in status, that sort of doesn't matter much for this test anyway

all we care about is that removal component removes generated secret, but keeps the other

tkdchen (approved):

mmorhun (approved):

PR #230, opened 2023-12-12
STONEBLD-2038 upgrade Tekton API

stuartwdouglas:
Upgrade the Tekton API to v0.50.5. Unfortunatly upstream has made changes to the types so that they no longer have JSON tags and as such are no longer compatible with controller-gen. To work around this I have introduced a parrallel set of types that have the JSON tags.


mmorhun:
Tekton API is already updated to `v0.57.0`. Closing as obsolete.

PR #233, opened 2023-12-14
STONEBLD-1844 follow on to run tasks in user namespace

stuartwdouglas:
Follow on from [STONEBLD-1844](https://issues.redhat.com//browse/STONEBLD-1844) to run the renovate job in the user namespace with very limited installation token.

### DIFF
--- a/controllers/component_dependency_update_controller.go
+++ b/controllers/component_dependency_update_controller.go
@@ -68,6 +68,10 @@ const (
 	FailureRetryTime = time.Minute * 5 // We retry after 5 minutes on failure
 )
 
+// The amount of time we wait before attempting to update the component, to try and avoid contention issues
+// This is not a constant so tests don't have to wait
+var delayTime = time.Second * 10
+
 // ComponentDependencyUpdateReconciler reconciles a PipelineRun object
 type ComponentDependencyUpdateReconciler struct {
 	client.Client
@@ -153,6 +157,10 @@ func (r *ComponentDependencyUpdateReconciler) Reconcile(ctx context.Context, req
 		log.Error(err, "Failed to get pipelineRun")
 		return ctrl.Result{}, err
 	}
+	if pipelineRun.CreationTimestamp.Add(delayTime).After(time.Now()) {
+		// These objects are super contested at creation, we just wait 10s before attempting anything
+		return ctrl.Result{RequeueAfter: delayTime}, nil
+	}
 
 	component, err := GetComponentFromPipelineRun(r.Client, ctx, pipelineRun)
 	if err != nil || component == nil {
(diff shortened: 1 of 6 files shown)


controllers/component_dependency_update_controller.go
```diff
@@ -153,6 +155,11 @@ func (r *ComponentDependencyUpdateReconciler) Reconcile(ctx context.Context, req
 		log.Error(err, "Failed to get pipelineRun")
 		return ctrl.Result{}, err
 	}
+	if pipelineRun.CreationTimestamp.Add(delayTime).After(time.Now()) {
+		//these objects are super contested at creation
```
mmorhun, line 159: nitpick: comment style.

controllers/renovate_util.go
```diff
@@ -311,3 +312,138 @@ func CreateRenovaterJob(ctx context.Context, client client.Client, scheme *runti
 
 	return nil
 }
+
+func CreateRenovaterPipeline(ctx context.Context, client client.Client, scheme *runtime.Scheme, namespace string, installations []installationStruct, slug string, debug bool, js func(slug string, repositories []renovateRepository, info interface{}) (string, error), info interface{}) error {
```
mmorhun, line 316: Please add description.

controllers/component_dependency_update_controller.go
```diff
@@ -424,7 +437,7 @@ func DefaultDependenciesUpdate(ctx context.Context, client client.Client, scheme
 		return false, err
 	}
 	log.Info("creating renovate job")
-	err = CreateRenovaterJob(ctx, client, scheme, installationsToUpdate, slug, true, generateRenovateConfigForNudge, result)
+	err = CreateRenovaterPipeline(ctx, client, scheme, result.Component.Namespace, installationsToUpdate, slug, true, generateRenovateConfigForNudge, result)
```
mmorhun, line 440: Why pipeline and not a job?

stuartwdouglas: - Job's are removed after a timeout, so lots of nudges in a short period could make the namespace unusable due to pod Quota, while pipelines are pruned much more aggressively
- Users can view the results of pipelines and the results are stored, making debugging much easier
- Tekton automatically provides docker config from linked service accounts for private images, with a job I would need to implement this manually

stuartwdouglas: I am not sure if we should change the other job to also use a pipeline, but as we don't really have e2e tests I don't really want to make major changes to that code just before I go on PTO.

pkg/git/github/github_app.go
```diff
@@ -323,6 +323,15 @@ func getAppInstallationsForRepository(githubAppIdStr string, appPrivateKeyPem []
...
 
 	repoStruct, _, err := installationClient.client.Repositories.Get(context.TODO(), owner, repo)
+	if err != nil {
+		return nil, "", err
+	}
+	// Create a new token, that is only valid for this repo
+	token, _, err = client.Apps.CreateInstallationToken(
+		context.Background(),
+		*val.ID,
+		&github.InstallationTokenOptions{RepositoryIDs: []int64{*repoStruct.ID}})
```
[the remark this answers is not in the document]
stuartwdouglas: I don't think so, I tried this initially but the app token does not have permission to list the repository id. I don't want to mount a generic token in the users namespace for security reasons (e.g. if build-service had a RHTAP namespace, the initial token would also allow is to push to integration-service).

controllers/component_dependency_update_controller.go
```diff
@@ -167,31 +175,29 @@ func (r *ComponentDependencyUpdateReconciler) Reconcile(ctx context.Context, req
...
+		if err != nil || result != nil {
+			return *result, err
 		}
+		log.Info("adding finalizer for component nudge")
 		// We add a finalizer to make sure we see the run before it is deleted
 		// As tekton results should aggressivly delete when pruning is enabled
+		patch := client.MergeFrom(pipelineRun.DeepCopy())
 		controllerutil.AddFinalizer(pipelineRun, NudgeFinalizer)
-		err = r.Client.Update(ctx, pipelineRun)
+		err = r.Client.Patch(ctx, pipelineRun, patch)
```
chmeliik, line 200: If I'm reading it right, the `patch` is a deepcopy of the original pipelineRun object, without the finalizer. And then the code adds the finalizer to the `pipelineRun` object, and then calls `Patch()`. How does that work? I would expect the finalizer to be on the patch, not on the object we're patching

chmeliik: Ohh, it's flipped for MergePatch https://github.com/kubernetes-sigs/controller-runtime/blob/1b80b9629af8ec55ae200e2c649a9603533eed20/pkg/client/patch.go#L100-L102

stuartwdouglas:
The test failure is not related to the PR, it is a codecov upload issue.

chmeliik (commented):
Are the two changes in this commit coupled to each other? If not, could you split it into two commits, where one replaces `Update()` with `Patch()` and the other changes the Renovate job/pipeline to run in the user's namespace?

controllers/renovate_util.go
```diff
@@ -311,3 +313,146 @@ func CreateRenovaterJob(ctx context.Context, client client.Client, scheme *runti
...
+									{
+										Prefix: "TOKEN_",
+										SecretRef: &v1.SecretEnvSource{
+											LocalObjectReference: v1.LocalObjectReference{
+												Name: name,
+											},
+										},
+									},
+								},
+								Command: []string{"bash", "-c", strings.Join(renovateCmds, "; ")},
```
chmeliik, line 397: Does joining multiple `renovate` commands with `;` potentially hide failures?

controllers/renovate_util.go
```diff
@@ -311,3 +313,146 @@ func CreateRenovaterJob(ctx context.Context, client client.Client, scheme *runti
...
+	}
+	if err := client.Create(ctx, configMap); err != nil {
+		return err
+	}
+	if err := client.Create(ctx, pipelineRun); err != nil {
+		return err
+	}
+	log.Info(fmt.Sprintf("Pipeline %s triggered", pipelineRun.Name), logs.Action, logs.ActionAdd)
+	if err := controllerutil.SetOwnerReference(pipelineRun, secret, scheme); err != nil {
+		return err
```
chmeliik, line 444: I guess it's not possible to set the owner reference before `Create()`-ing the objects? Which means the secrets and configmaps may persist in the user namespace indefinitely (if e.g. the pipelineRun creation fails)?

controllers/renovate_util.go
stuartwdouglas: This is how the existing renovate job works. Also even if one fails we want the others to continue. The lack of a PR will make it clear that it has failed, and with this new approach the logs should be accessible.

controllers/renovate_util.go
stuartwdouglas: Actually it should be possible, because the PipelineRun will wait for creation, this is how the multi platform controller works. 

This code was copy pasted from the existing renovate code

mmorhun (approved):

controllers/component_dependency_update_controller.go
mmorhun: I see, thanks

stuartwdouglas:
Failures have been due to the corresponding e2e-tests branch needed to be rebased on main.

PR #235, opened 2023-12-18
fix(rhtapbugs=975): use source_url instead of repo_url for git-url

14rcole:
The repo_url variable in pull request jobs points to the target repo. The git-url parameter of build pipelineRuns is passed to the integration service.  If users attempt to clone this URL with the revision from the 'revision' parameter (which is the revision of the source URL) an error will occur.  Tekton also has a 'source_url' variable that points to the source repo in pull request jobs, which fixes the problem.  In push jobs 'source_repo' defaults to the same value as 'repo_url'.  Therefore defaulting to 'source_repo' should not cause any other bugs'

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/controllers/component_build_controller_pac.go
+++ b/controllers/component_build_controller_pac.go
@@ -1076,7 +1076,7 @@ func generatePaCPipelineRunForComponent(
 	}
 
 	params := []tektonapi.Param{
-		{Name: "git-url", Value: tektonapi.ParamValue{Type: "string", StringVal: "{{repo_url}}"}},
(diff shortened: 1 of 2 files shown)


mmorhun (approved):
@14rcole thank you for the improvement.
LGTM, but please make sure that e2e tests are passing.

14rcole:
@mmorhun Can you please merge this?  I don't have write access.

PR #240, opened 2024-01-05
RHTAP-1134: Support custom Gitlab instances

gbenhaim:
When creating a Gitlab client, the base url of the Gitlab instance should be provided (unless the default is used, which is https://gitlab.com). With this change, it would be possible to use Gitlab instance other than gitlab.com.

In addition, fixed an issue in "getWebhookByTargetUrl" where it tried to access a nil pointer of a request.

### DIFF
--- a/pkg/git/gitlab/gitlab_helper.go
+++ b/pkg/git/gitlab/gitlab_helper.go
@@ -30,6 +30,31 @@ import (
 	"net/url"
 )
 
+type FailedToParseUrlError struct {
+	url string
+	err string
+}
+
+func (e FailedToParseUrlError) Error() string {
+	return fmt.Sprintf("Failed to parse url: %s, error: %s", e.url, e.err)
+}
+
+type MissingSchemaError struct {
+	url string
+}
+
+func (e MissingSchemaError) Error() string {
+	return fmt.Sprintf("Failed to detect schema in url %s", e.url)
+}
+
+type MissingHostError struct {
+	url string
+}
+
+func (e MissingHostError) Error() string {
+	return fmt.Sprintf("Failed to detect host in url %s", e.url)
+}
+
 func getProjectPathFromRepoUrl(repoUrl string) (string, error) {
 	url, err := url.Parse(repoUrl)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/git/gitlab/gitlab_helper_test.go
```diff
@@ -56,3 +56,123 @@ func TestGetProjectPathFromRepoUrl(t *testing.T) {
...
 }
+
+func TestGetHostFromRepoUrl(t *testing.T) {
+	var params = []struct {
+		in  string
+		out string
+		err error
+	}{
+		{
+			"https://gitlab.cee.redhat.com/rhtap/tenants-config.git",
```
chmeliik, line 67: AFAIK we should generally avoid mentioning internal URLs in upstream code, could you replace the redhat.com urls with something else?

chmeliik (commented):
LGTM with a minor nitpick for the test data

pkg/git/gitlab/gitlab_helper_test.go
gbenhaim: done

pkg/git/gitlab/gitlab_helper.go
```diff
@@ -42,6 +67,24 @@ func getProjectPathFromRepoUrl(repoUrl string) (string, error) {
 	), nil
 }
 
+func GetBaseUrl(repoUrl string) (string, error) {
+	url, err := url.Parse(repoUrl)
+	if err != nil {
+		return "", FailedToParseUrlError{url: repoUrl, err: err.Error()}
```
mmorhun, line 73: Custom errors looks redundant here: they are never used / distinguished.

mmorhun (approved):
LGTM in general, just a small remark.
Thanks for contributions!

pkg/git/gitlab/gitlab_helper.go
gbenhaim: they are used in the tests :)

gbenhaim:
@chmeliik @mmorhun can someone please merge it?

gbenhaim:
@chmeliik can you please merge it? I see that all the checks passed. thanks!

PR #243, opened 2024-01-22
Allow renovatebot to rebase existing PRs on staging after renaming

mmorhun:
After renaming of renovate bot name, existing PRs on staging are not getting updates. This PR adds old bot name into exceptions to overwrite changes from, so exiting PRs could be updated.
Needs to be reverted after some time.

### DIFF
--- a/controllers/git_tekton_resources_renovater.go
+++ b/controllers/git_tekton_resources_renovater.go
@@ -172,6 +172,7 @@ func generateConfigJS(slug string, repositories []renovateRepository, _ interfac
 				recreateClosed: true,
 				recreateWhen: "always",
 				rebaseWhen: "behind-base-branch",
+				gitIgnoredAuthors: ["rhtap-staging"],
 				enabled: true
 			  }
 			]


controllers/git_tekton_resources_renovater.go
```diff
@@ -172,6 +172,7 @@ func generateConfigJS(slug string, repositories []renovateRepository, _ interfac
 				recreateClosed: true,
 				recreateWhen: "always",
 				rebaseWhen: "behind-base-branch",
+				gitIgnoredAuthors: ["rhtap-staging"],
```
chmeliik, line 175: Any reason not to add the prod name as well?

mmorhun: Prod one wasn't renamed

chmeliik: But if there are plans to rename it as well, won't it break the same way it did in stage?

mmorhun: I\m not aware of such plans. If that decided, we can add one more exception.

chmeliik (approved):

lcarva (approved):

tkdchen (approved):

PR #246, opened 2024-01-29
Fix kube linter

stuartwdouglas:


tkdchen (approved):

mmorhun:
@stuartwdouglas @tkdchen for what specific places this exclusion is made?
I'm afraid, kubelinter will fail in infra-deployments.

PR #251, opened 2024-02-16
Enable the customization of nudged files

arewm:
Fixes [KFLUXBUGS-1114](https://issues.redhat.com/browse/KFLUXBUGS-1114)

To prevent us from having to continuously modify the set of files that can have nudged references, we will define the default set of files and let users customize the nudges based on pipeline parameters. There is no intention to add default values or to expose this parameter to users in the PipelineDefinition pushed out to repositories.

I chose to add these to the Tekton params even though the params are not being used by any task in the pipeline. If there is opposition to this, I can probably reimplement it as setting the parameter by annotations.

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/controllers/component_dependency_update_controller.go
+++ b/controllers/component_dependency_update_controller.go
@@ -498,6 +498,14 @@ func generateRenovateConfigForNudge(slug string, repositories []renovateReposito
 	buildResult := context.(*BuildResult)
 
 	repositoriesData, _ := json.Marshal(repositories)
+	fileMatchParts := strings.Split(buildResult.FileMatches, ",")
+	for i := range fileMatchParts {
+		fileMatchParts[i] = strings.TrimSpace(fileMatchParts[i])
+	}
+	fileMatch, err := json.Marshal(fileMatchParts)
+	if err != nil {
+		return "", err
+	}
 	body := `
 	{{with $root := .}}
 	module.exports = {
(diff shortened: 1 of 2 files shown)


controllers/component_dependency_update_controller.go
```diff
@@ -500,7 +514,7 @@ func generateRenovateConfigForNudge(slug string, repositories []renovateReposito
     	enabledManagers: "regex",
 		customManagers: [
 			{
-            	"fileMatch": [".*Dockerfile.*",".*.yaml",".*Containerfile.*"],
+            	"fileMatch": [{{range .FileMatches}},"{{.}}"{{end}}],
```
arewm, line 517: Do we need to escape or sanitize the input `FileMatches` before generating this config? I think so, but what escaping should be done?

stuartwdouglas: I don't know if we need to worry about that for now. If you have permission to modify this then you already have permission to run a pipeline, so I don't think there is a security aspect to it, and there are no real valid uses casess where escaping would be required.

stuartwdouglas (approved):

controllers/component_dependency_update_controller.go
```diff
@@ -264,6 +266,16 @@ func (r *ComponentDependencyUpdateReconciler) handleCompletedBuild(ctx context.C
 		repo = image[0:index]
 		tag = image[index+1:]
 	}
+	// find any configurations for files to nudge in
+	nudgeFiles := []string{}
+	for _, p := range pipelineRun.Spec.Params {
```
stuartwdouglas, line 271: Thinking about it a bit is a parameter really the best place for this? Wouldn't an annotation make more senses?

stuartwdouglas (changes requested):

controllers/component_dependency_update_controller.go
chmeliik: Maybe we can just JSON-encode the strings? Are JSON strings valid JS strings?

stuartwdouglas (approved):

PR #253, opened 2024-02-28
KFLUXBUGS-1113 strip .git from URL

stuartwdouglas:


controllers/renovate_util.go
```diff
@@ -136,7 +136,8 @@ func GetGithubInstallationsForComponents(ctx context.Context, client client.Clie
 
 		gitSource := component.Spec.Source.GitSource
 
-		githubAppInstallation, slugTmp, err := github.GetAppInstallationsForRepository(githubAppIdStr, privateKey, gitSource.URL)
+		url := strings.TrimSuffix(strings.TrimSuffix(gitSource.URL, ".git"), "/")
```
chmeliik, line 139: Can git urls end with `.git/` and `/.git`? Currently this handles only the latter case

stuartwdouglas: It was copied from the existing renovate code for the RHTAP updates, so I don't think it is a real issue. If you copy from the browser address you might get a / or from the github clone url you would get a .git, but that kind of messed up URL you would need to manually enter.

chmeliik: Ok, better to keep it consistent within the codebase

PR #255, opened 2024-03-08
When github application is suspended don't treat it as transient error

rcerven:
[STONEBLD-2218](https://issues.redhat.com//browse/STONEBLD-2218)

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/pkg/boerrors/perror.go
+++ b/pkg/boerrors/perror.go
@@ -108,6 +108,8 @@ const (
 	// GitHub Application with specified ID does not exists.
 	// Correct configuration in the AppStudio installation ('pipelines-as-code-secret' secret in 'build-service' namespace).
 	EGitHubAppDoesNotExist BOErrorId = 73
+	// EGitHubAppSuspended Application in git repository is suspended
+	EGitHubAppSuspended BOErrorId = 78
 
 	// EGitHubTokenUnauthorized access token can't be recognized by GitHub and 401 is responded.
 	// This error may be caused by a malformed token string or an expired token.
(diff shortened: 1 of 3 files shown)


controllers/component_build_controller_pac.go
```diff
@@ -871,6 +871,11 @@ func (r *ComponentBuildReconciler) ConfigureRepositoryForPaC(ctx context.Context
 		IsAppInstallationExpected: true,
 	})
 	if err != nil {
+		if strings.Contains(err.Error(), "installation has been suspended") {
+			log.Info(fmt.Sprintf("Installation has been suspended for git repository: %s", repoUrl))
```
mmorhun, line 875: Logs reader is not in context of the code, so please describe installation of what and where has been suspended. And probably consequences for the Component.

controllers/component_build_controller_pac.go
```diff
@@ -871,6 +871,11 @@ func (r *ComponentBuildReconciler) ConfigureRepositoryForPaC(ctx context.Context
 		IsAppInstallationExpected: true,
 	})
 	if err != nil {
+		if strings.Contains(err.Error(), "installation has been suspended") {
```
mmorhun, line 874: Does it make sense to check the condition inside the git package like for other errors, e.g. [this](https://github.com/redhat-appstudio/build-service/blob/main/pkg/git/github/github_app.go#L61)?

rcerven: 403 is general forbidden, so I'd rather check error message

mmorhun: Yes, we definitely need to check error message, my point was in the place of the check.

rcerven: ok I'll check in github package

mmorhun (approved):

PR #257, opened 2024-03-14
RHTAPSRE-469: Webhook configuration based on repo url

gbenhaim:
Allow to configure the webhook target url based on the repository url of the component by providing a config file.

Configuring the webhook target url using environment variable has precedence.

In case that no config file or environment variable for configuring the webhook target are provided, the PAC route will be used (same as before this change).

An example use case would be for a cluster that sits behind a firewall, and need to receive webhooks from SCMs outside of the firewall (through a relay service) and also from SCMs within the firewall, in this case the webhooks can be emitted directly to the PAC endpoint on the cluster, instead of the relay service.

### DIFF
--- a/controllers/component_build_controller.go
+++ b/controllers/component_build_controller.go
@@ -40,6 +40,7 @@ import (
 	appstudiov1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/build-service/pkg/boerrors"
 	l "github.com/redhat-appstudio/build-service/pkg/logs"
+	"github.com/redhat-appstudio/build-service/pkg/webhook"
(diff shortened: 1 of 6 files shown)


mmorhun:
@gbenhaim could you also explain in the PR description for which use cases this functionality is needed?

gbenhaim:
@mmorhun done

mmorhun (approved):
You'll have to reference changes from https://github.com/redhat-appstudio/application-service/pull/450 before merging

gbenhaim:
> You'll have to reference changes from [redhat-appstudio/application-service#450](https://github.com/redhat-appstudio/application-service/pull/450) before merging

Those two PRs are not dependent.

PR #268, opened 2024-04-08
Don't allow HTTP protocol for git repo url

rcerven:
[STONEBLD-2301](https://issues.redhat.com//browse/STONEBLD-2301)

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually


tkdchen (commented):
Maybe, the source URL can be checked inside the `Reconcile` when all criteria are met and begin to handle the build with component, that can reduce duplicate code a little.

rcerven:
> Maybe, the source URL can be checked inside the `Reconcile` when all criteria are met and begin to handle the build with component, that can reduce duplicate code a little.

yes I thought about it, but I'd have to add additional logic to main reconcile and setup "some" status on component etc

and also I don't won't to block for example unprovision

tkdchen (approved):

mmorhun (approved):

PR #284, opened 2024-04-26
Respect self-hosted GitLab URL in Repository CR

mmorhun:
In case of OnPrem GitLab instance, Repository CR should respect the instance URL.


rcerven (approved):

gabemontero:
@mmorhun - either @enarha @savitaashture or @chmouel would be better reviewers of this change than me.  I have rudimentary PAC skills at best.

PR #287, opened 2024-05-02
test: What happens when a Go module moves to a different repo?

chmeliik:


chmeliik:
Testing dependency on:
* github.com/chmeliik/some-go-lib @ [a51b711](https://github.com/acmiel-rhtap/some-go-lib/commit/a51b711113bcc68bc088432a0949175215c91fd1)
  * has moved to github.com/acmiel-rhtap/some-go-lib
  * but github sets up an automatic redirect, and at that commit the name in `go.mod` was still matching
  * so it's still possible to download and import this dependency
* github.com/acmiel-rhtap/some-go-lib @ [c756c7b](https://github.com/acmiel-rhtap/some-go-lib/commit/c756c7b9c8fb7a6314b0f704c554737b692def5c)
  * can't use commit a51b711, because the name in `go.mod` doesn't match
  * after updating to the new commit, which updates the name in go.mod to match the repo path, it works

TLDR: moving a Go module to a different repo doesn't break those who depend on the module :exploding_head: 
(at least on Github, where you get automatic redirects)

PR #289, opened 2024-05-03
Move repo to konflux-ci org

chmeliik:
[STONEBLD-2336](https://issues.redhat.com//browse/STONEBLD-2336)

Signed-off-by: Adam Cmiel <acmiel@redhat.com>

PR #290, opened 2024-05-03
nudge controller: fix broken renovate config

chmeliik:
When there is more than 1 distribution repository, the controller
generates a syntactically invalid javascript file. Add the missing
comma between key-value pairs.

Signed-off-by: Adam Cmiel <acmiel@redhat.com>


arewm (approved):

chmeliik:
Btw we probably need to merge #289 first and then rebase. Until then, the push pipeline will fail to update infra-deployments (the org has changed)

rcerven (approved):

PR #316, opened 2024-06-20
Remove renovate service that updates user's pipelines definition from Build Service operator

mmorhun:
Since Konflux is switching to new Mintmaker service that covers update of user's pipeline definitions and more, we retiring renovate controller from Build Service.

Note, renovate for component nudging is kept and will not be replaced/retired.


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Build the manager binary
 # For more details and updates, refer to
 # https://catalog.redhat.com/software/containers/ubi9/go-toolset/61e5c00b4ec9945c18787690
-FROM registry.access.redhat.com/ubi9/go-toolset:1.20.10 AS builder
```
MartinBasti, line 4: Why go version bump is needed for removal of a feaure?

mmorhun: Not really needed, but since some operators are already updated, we need to update build service too.
I agree, that it would be good to have a separate PR for that, but since we are time constrained, I did it on the go.

MartinBasti: I fully disagree that time constrains are reason to mix multiple things into one commit

rcerven (approved):

PR #329, opened 2024-07-19
remove simple build feature

rcerven:
[STONEBLD-2612](https://issues.redhat.com//browse/STONEBLD-2612)

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually


mmorhun (commented):
LGTM in general.
The only thing is that we still have `component_build_controller_common.go` file for build controller, but with the removal of simple build, it's not a thing any more. So, ideally, we shouldn't have it. The simplest way is to move all its content to PaC file, but it's already overloaded with 1500+ lines. We need some refactor here.
It's ok if we do it in follow up work, but up to you.

mmorhun (approved):
Let's have the refactor in the follow up PR

PR #331, opened 2024-07-26
nudging components now works for multiple repositories in the same

rcerven:
installation

[STONEBLD-2647](https://issues.redhat.com//browse/STONEBLD-2647)

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/controllers/renovate_util.go
+++ b/controllers/renovate_util.go
@@ -32,6 +32,7 @@ const (
 
 type installationStruct struct {
 	id           int
+	component    string
 	token        string
 	repositories []renovateRepository
 }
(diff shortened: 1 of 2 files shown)


arewm (approved):
Thanks for all of the extra logging as well!

controllers/renovate_util.go
```diff
@@ -32,6 +32,7 @@ const (
 
 type installationStruct struct {
 	id           int
+	component    string
 	token        string
 	repositories []renovateRepository
 }
```
chmeliik, line 38: Hmm, so what is an "installation"? Why does it have just one `component` but a list of `repositories`?

rcerven: well it isn't exactly like WHOLE installation struct

but installation for component (1 repository) struct


so 1 struct has info about installation and specific repository for component 

and list of repositories has in fact only 1 repository ......

PR #340, opened 2024-08-21
Improve the heuristic for guessing the git provider

chmeliik:
[KFLUXBUGS-1559](https://issues.redhat.com//browse/KFLUXBUGS-1559)

Previously, the function guessed the provider based on a specific
segment in the hostname in the repo url.

For a host like cooler.gitlab.my-company.com, the returned provider
would be (depending on whether the repo url is git@ -style or not):

    git@cooler.gitlab.my-company.com:org/repo
        ^^^^^^

    https://cooler.gitlab.my-company.com/org/repo.git
                          ^^^^^^^^^^

Fix the heuristic and make it much looser - now it just searches the
hostname for substrings corresponding to the known git providers.

Note: in the highly unusual case where a hostname contains substrings
corresponding to more than one provider, the precedence is github >
gitlab > bitbucket. (Making the precedence deterministic is the reason
why allowedGitProviders had to change from a map to a slice).

### Checklist:
 - Update readme and documentation - TODO (internal documentation only)

### DIFF
--- a/controllers/component_build_controller_pac.go
+++ b/controllers/component_build_controller_pac.go
@@ -607,40 +607,39 @@ func (r *ComponentBuildReconciler) UnconfigureRepositoryForPaC(ctx context.Conte
 // getGitProvider returns git provider name based on the repository url or the git-provider annotation
 func getGitProvider(component appstudiov1alpha1.Component) (string, error) {
 	allowedGitProviders := []string{"github", "gitlab", "bitbucket"}
-	gitProvider := ""
 
-	if component.Spec.Source.GitSource == nil {
-		err := fmt.Errorf("git source URL is not set for %s Component in %s namespace", component.Name, component.Namespace)
-		return "", err
-	}
-
-	sourceUrl := component.Spec.Source.GitSource.URL
-	var host string
-
-	if strings.HasPrefix(sourceUrl, "git@") {
-		// git@github.com:redhat-appstudio/application-service.git
-		sourceUrl = strings.TrimPrefix(sourceUrl, "git@")
-		host = strings.Split(sourceUrl, ":")[0]
-	} else {
-		// https://github.com/redhat-appstudio/application-service
-		u, err := url.Parse(sourceUrl)
-		if err != nil {
-			return "", err
-		}
-		host = u.Hostname()
-	}
-
-	for _, provider := range allowedGitProviders {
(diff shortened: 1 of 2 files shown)


tisutisu:
[moving to bigger instance](https://github.com/openshift/release/pull/56019) change is merged
/test build-service-e2e

tisutisu:
Haven't used changes, so retesting
/test build-service-e2e

mmorhun (approved):

controllers/component_build_controller_pac.go
```diff
@@ -603,50 +604,49 @@ func (r *ComponentBuildReconciler) UnconfigureRepositoryForPaC(ctx context.Conte
...
-	allowedGitProviders := map[string]bool{"github": true, "gitlab": true, "bitbucket": true}
+	allowedGitProviders := []string{"github", "gitlab", "bitbucket"}
 	gitProvider := ""
 
 	if component.Spec.Source.GitSource == nil {
 		err := fmt.Errorf("git source URL is not set for %s Component in %s namespace", component.Name, component.Namespace)
 		return "", err
 	}
+
+	gitProvider = component.GetAnnotations()[GitProviderAnnotationName]
```
mmorhun, line 617: nitpick: `gitProvider` can be defined here.

controllers/component_build_controller_pac.go
```diff
@@ -603,50 +604,49 @@ func (r *ComponentBuildReconciler) UnconfigureRepositoryForPaC(ctx context.Conte
...
+	if gitProvider != "" && !slices.Contains(allowedGitProviders, gitProvider) {
```
mmorhun, line 618: nitpick: if the annotation has value we just use it (or error) before applying any heuristics. IMO, that would make the code easier to read.

chmeliik: That was my intention, the second commit makes that change. PTAL if that's what you meant

chmeliik: Or did you mean that we should `return` right away from the function instead of using `if gitProvider == ""` below?

chmeliik: > Or did you mean that we should return right away from the function instead of using if gitProvider == "" below?

FWIW, I tried that and it didn't seem more readable than the current code

PR #341, opened 2024-08-21
incoming webhook trigger doesn't work with gitlab,

rcerven:
Repository CR, needs spec->git_provider->type specified for anything what isn't github app

[STONEBLD-2681](https://issues.redhat.com//browse/STONEBLD-2681)

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/controllers/component_build_controller_pac_repository.go
+++ b/controllers/component_build_controller_pac_repository.go
@@ -225,6 +225,15 @@ func generatePACRepository(component appstudiov1alpha1.Component, config *corev1
 		gitProviderConfig.URL = url
 	}
 
+	if gitProviderConfig != nil {
+		repositoryGitProvider := gitProvider
+		// https://pipelinesascode.com/docs/guide/incoming_webhook/#incoming-webhook-url
+		if repositoryGitProvider == "bitbucket" {
+			repositoryGitProvider = "bitbucket-cloud"
+		}
+		gitProviderConfig.Type = repositoryGitProvider
+	}
+
 	repository := &pacv1alpha1.Repository{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Repository",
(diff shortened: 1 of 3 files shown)


controllers/component_build_controller_pac_repository.go
```diff
@@ -225,6 +225,14 @@ func generatePACRepository(component appstudiov1alpha1.Component, config *corev1
 		gitProviderConfig.URL = url
 	}
 
+	if gitProviderConfig != nil {
+		repositoryGitProvider := gitProvider
+		if repositoryGitProvider == "bitbucket" {
+			repositoryGitProvider = "bitbucket-cloud"
```
chmeliik, line 231: Could you add a comment with a link to some documentation that would explain why `bitbucket-cloud` is the correct value?

controllers/component_build_controller_unit_test.go
chmeliik, line 1: It looks like we don't have a bitbucket test case

controllers/component_build_controller_pac_repository.go
rcerven: added

chmeliik: nitpick: https://pipelinesascode.com/docs/guide/incoming_webhook/#incoming-webhook-url looks like the more relevant section

controllers/component_build_controller_unit_test.go
chmeliik: Discussed offline: bitbucket isn't fully supported, no need to add a testcase yet

chmeliik (approved):

controllers/component_build_controller_pac_repository.go
rcerven: updated

mmorhun (approved):

controllers/component_build_controller_pac_repository.go
savitaashture: @rcerven naive query 
may be i m missing something 
PAC already support bitbucket-cloud as per the doc https://pipelinesascode.com/docs/guide/incoming_webhook/#incoming-webhook-url any reason for overriding it :thinking:

rcerven: @savitaashture PAC requires 'bitbucket-cloud' string, but in build-service we are using 'bitbucket' that is why I am converting it to string which PAC understands

savitaashture: Okay got it thank you

PR #344, opened 2024-09-05
Do not recreate pipeline service account on component deletetion

mmorhun:
Build Service should not attempt to recreate build pipeline service account on component deletion. This matters in case of whole namespace removal, so such attempt would fail blocking finalizers processing.
Also includes small code corrections.

### DIFF
--- a/controllers/component_build_controller_pipeline.go
+++ b/controllers/component_build_controller_pipeline.go
@@ -467,10 +467,6 @@ func getContainerImageRepositoryForComponent(component *appstudiov1alpha1.Compon
 	if component.Spec.ContainerImage != "" {
 		return getContainerImageRepository(component.Spec.ContainerImage)
 	}
-	imageRepo, _, err := getComponentImageRepoAndSecretNameFromImageAnnotation(component)
-	if err == nil && imageRepo != "" {
-		return imageRepo
-	}
 	return ""
 }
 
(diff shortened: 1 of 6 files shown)


controllers/component_build_controller_test.go
```diff
@@ -522,18 +522,6 @@ var _ = Describe("Component initial build controller", func() {
 			expectPacBuildStatus(resourcePacPrepKey, "enabled", 0, "", mergeUrl)
 		})
 
-		It("should not copy PaC secret into local namespace if GitHub application is used", func() {
```
tnevrlka, line 525: Wondering, why is this getting deleted?

controllers/suite_util_test.go
```diff
@@ -599,3 +599,45 @@ func deleteBuildPipelineConfigMap(configMapKey types.NamespacedName) {
 		Fail(err.Error())
 	}
 }
+
+func waitServiceAccount(serviceAccountKey types.NamespacedName) {
```
tnevrlka, line 603: Could we improve this function name? Maybe something like `waitForServiceAccount()`?

controllers/component_build_controller_pipeline.go
```diff
@@ -484,25 +480,6 @@ func getContainerImageRepository(image string) string {
 	return strings.Split(image, ":")[0]
 }
 
-// getComponentImageRepoAndSecretNameFromImageAnnotation parses image.redhat.com/image annotation
-// for image repository and secret name to access it.
-// If image.redhat.com/image is not set, the procedure returns empty values.
-func getComponentImageRepoAndSecretNameFromImageAnnotation(component *appstudiov1alpha1.Component) (string, string, error) {
```
tnevrlka, line 490: Wondering, is this no longer used?

controllers/component_build_controller_test.go
mmorhun: That functionality is removed. Build Service is not copying any secret any more.

controllers/suite_util_test.go
mmorhun: I followed pattern used, see other functions with `wait` prefix, like `waitSecretGone`, etc.

controllers/component_build_controller_pipeline.go
mmorhun: That was a fallback, but that option is deleted with old "annotation" reconciler in [Image Controller](https://github.com/konflux-ci/image-controller).

tnevrlka (approved):

tnevrlka (approved):
/lgtm

mmorhun:
First time e2e tests passed. In the second commit I've deleted unused function because of linter complain.
e2e tests failure is just tests instability, so merging this PR.

PR #369, opened 2024-11-12
when POST fails, buils-service starts another reconcile, but PaC

rcerven:
actually triggers new build, resulting in endless pipelines

temporary workaround for KFLUXBUGS-1858

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/controllers/component_build_controller_pac.go
+++ b/controllers/component_build_controller_pac.go
@@ -299,7 +299,8 @@ func (r *ComponentBuildReconciler) TriggerPaCBuild(ctx context.Context, componen
 	resp, err := HttpClient.Post(triggerURL, "application/json", bytes.NewBuffer(bytesParam))
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mmorhun (approved):

arewm (approved):
Will this result in the status updating on the Component? 

Will this be user facing other than not having many PLRs being triggered?

What process would be required if the sync between build service and PAC legitimately fails? How would a user/SRE notice and recover?

rcerven:
> Will this result in the status updating on the Component?
> 
yes it will be processed as successful and annotation updated

> Will this be user facing other than not having many PLRs being triggered?
> 
well PaC actually triggers the pipeline

> What process would be required if the sync between build service and PAC legitimately fails? How would a user/SRE notice and recover?
just from build-service logs can error be seen

PR #384, opened 2024-12-16
Ensure protocol in git provider URL in Repository CR

mmorhun:
Users often provide only hostname for self-hosted git providers. Make sure that git provider URL in Repository CR has protocol set, otherwise PaC would fail.
Refactor code to make it more readable.


controllers/component_build_controller_unit_test.go
```diff
@@ -1019,6 +1019,65 @@ func TestGeneratePACRepository(t *testing.T) {
...
+				URL:  "https://github.com",
+				Type: "github",
+			},
+		},
+		{
+			name:    "should create PaC repository for GitHub application on self-hosted GitHub",
+			repoUrl: "https://github.self-hosted.com/user/test-component-repository",
+			componentAnnotations: map[string]string{
+				GitProviderAnnotationName: "github",
+				GitProviderAnnotationURL:  "https://github.self-hosted.com",
```
chmeliik, line 1031: Do we  have a test where the protocol is missing?

mmorhun: Good catch, now we do

chmeliik (approved):

rcerven (approved):

tisutisu:
/retest
cleaned up some old private repos

PR #392, opened 2025-01-22
manually update konflux references

tnevrlka:
Enterprise Contract checks are currently failing due to outdated task versions.
There is a [Mintmaker PR open](https://github.com/konflux-ci/build-service/pull/389), but it does not include the latest fix for CycloneDX version and the PR will not get updated until Saturday.

Manually update to the references provided by Mintmaker, but also include the newer buildah task version which has the fix for CycloneDX version.

### DIFF
--- a/.tekton/build-service-pull-request.yaml
+++ b/.tekton/build-service-pull-request.yaml
@@ -134,7 +134,7 @@ spec:
         - name: name
           value: init
         - name: bundle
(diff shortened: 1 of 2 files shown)


tnevrlka:
My bad, seemed to me like the original Mintmaker PR wouldn't get updated before Saturday, but it did in the end. Closing in favor of https://github.com/konflux-ci/build-service/pull/389

PR #413, opened 2025-02-28
fix: do not store SL token

psturc:
### JIRA
https://issues.redhat.com/browse/KFLUXDP-180

### Changes
* adding a new parameter to the Sealights instrumentation task to skip saving the token in TA - that way it is safe to push the image to public repository


psturc:
> Should we also push the build-service image itself into the private repo? Or is it enough to push the source-code TA and the prefetch TA to the private repo?

So we decided to change the approach and use the flag `disable-token-save` for not storing the token in the artifact. Then both TA and image should be safe to store in public repo.

PR #417, opened 2025-03-12
when git source url has invalid host or not existing repo it is perma…

rcerven:
…nent error

[STONEBLD-3272](https://issues.redhat.com//browse/STONEBLD-3272)

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/internal/controller/component_build_controller_pac.go
+++ b/internal/controller/component_build_controller_pac.go
@@ -254,7 +254,7 @@ func (r *ComponentBuildReconciler) TriggerPaCBuild(ctx context.Context, componen
 	}
 
 	// getting branch in advance just to test credentials
-	defaultBranch, err := gitClient.GetDefaultBranch(repoUrl)
+	defaultBranch, err := gitClient.GetDefaultBranchWithChecks(repoUrl)
 	if err != nil {
 		return false, err
 	}
@@ -443,7 +443,7 @@ func (r *ComponentBuildReconciler) ConfigureRepositoryForPaC(ctx context.Context
 	}
 
 	// getting branch in advance just to test credentials
-	defaultBranch, err := gitClient.GetDefaultBranch(repoUrl)
+	defaultBranch, err := gitClient.GetDefaultBranchWithChecks(repoUrl)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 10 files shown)


internal/controller/component_build_controller_pac.go
```diff
@@ -715,3 +715,11 @@ func getGitProvider(component appstudiov1alpha1.Component) (string, error) {
 
 	return gitProvider, nil
 }
+
+// getGitUrlError returns more specific git error
+func getGitUrlError(err error) error {
```
mmorhun, line 720: Maybe `checkGitUrlError`?

mmorhun (commented):
This makes me think, that maybe we should introduce a new gitclient method, like `getDefaultBranchWithChecks` or something, since we already doing some check (for credentials).

rcerven:
> This makes me think, that maybe we should introduce a new gitclient method, like `getDefaultBranchWithChecks` or something, since we already doing some check (for credentials). WDYT?

reworked

pkg/git/gitprovider/common.go
```diff
@@ -33,3 +39,11 @@ func IsInsecureSSL() bool {
 	}
 	return false
 }
+
+// CheckGitUrlError returns more specific git error
+func CheckGitUrlError(err error) error {
```
mmorhun, line 44: This must be implemented in each git provider because:
1. We adding dependency for a specific internal package
2. Different providers might have different responses

pkg/git/github/github_client.go
```diff
@@ -291,10 +291,16 @@ func (g *GithubClient) DeletePaCWebhook(repoUrl, webhookUrl string) error {
...
-func (g *GithubClient) GetDefaultBranch(repoUrl string) (string, error) {
+// GetDefaultBranchWithChecks returns name of default branch in the given repository
+// also performs additional checks if repo exists and hostname is valid and returns more specific error
+func (g *GithubClient) GetDefaultBranchWithChecks(repoUrl string) (string, error) {
 	owner, repository := getOwnerAndRepoFromUrl(repoUrl)
-	return g.getDefaultBranch(owner, repository)
+
+	defaultBranch, err := g.getDefaultBranch(owner, repository)
+	if err != nil {
+		return "", gp.CheckGitUrlError(err)
```
mmorhun, line 301: Does it make sense to do the check in `refineGitHostingServiceError` ? Then `WithChecks` suffix can be omitted

pkg/git/gitprovider/common.go
rcerven: moved

mmorhun (approved):

PR #425, opened 2025-04-11
Create dedicated build SA per Component

mmorhun:
- Create a dedicated Service Account for each Component named `build-pipeline-<component-name>` to run build pipelines with.
- Create migration code that would create a PR into user's git repository that specify explicitly Service Account to use for the build pipeline. The migration will be performed per Component if `build.appstudio.openshift.io/sa-migration: 'true'` annotation is added to the Component.
- Build Service will continue creation of old style PaC provision PRs, until `use-new-sa` Config Map is created in `build-service` namespace. Then the new Service Account would be used by default.
- Introduce a new temporal controller that syncs linked secrets from deprecated `appstudio-pipeline` Service Account to the dedicated Component build Service Account. This is needed for UI until new functionality with dedictaed Service Accounts implemented.

### DIFF
--- a/internal/controller/appstudio-pipeline-sa-watcher-controller.go
+++ b/internal/controller/appstudio-pipeline-sa-watcher-controller.go
@@ -79,12 +79,21 @@ func (r *AppstudioPipelineServiceAccountWatcherReconciler) Reconcile(ctx context
 		return ctrl.Result{}, err
 	}
 
+	commonServiceAccount := &corev1.ServiceAccount{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: buildPipelineServiceAccountName, Namespace: req.Namespace}, commonServiceAccount); err != nil {
+		if errors.IsNotFound(err) {
+			// Assume the migration to the dedicated build Service Account has been done, nothing to do.
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
 	for _, component := range componentList.Items {
-		buildPipelineServiceAccountName := getBuildPipelineServiceAccountName(&component)
+		dedicatedBuildPipelineServiceAccountName := getBuildPipelineServiceAccountName(&component)
 		buildPipelinesServiceAccount := &corev1.ServiceAccount{}
-		if err := r.Client.Get(ctx, types.NamespacedName{Name: buildPipelineServiceAccountName, Namespace: req.Namespace}, buildPipelinesServiceAccount); err != nil {
+		if err := r.Client.Get(ctx, types.NamespacedName{Name: dedicatedBuildPipelineServiceAccountName, Namespace: req.Namespace}, buildPipelinesServiceAccount); err != nil {
 			if !errors.IsNotFound(err) {
-				log.Error(err, fmt.Sprintf("failed to read build pipeline Service Account %s in namespace %s", buildPipelineServiceAccountName, component.Namespace), l.Action, l.ActionView)
+				log.Error(err, fmt.Sprintf("failed to read build pipeline Service Account %s in namespace %s", dedicatedBuildPipelineServiceAccountName, component.Namespace), l.Action, l.ActionView)
 				// do not break sync because of a faulty item
 			}
 			// Dedicated build pipeline Service Account hasn't been yet created.
@@ -92,7 +101,7 @@ func (r *AppstudioPipelineServiceAccountWatcherReconciler) Reconcile(ctx context
 			continue
 		}
 
-		if err := LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx, r.Client, &component); err != nil {
+		if err := LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx, r.Client, &component, commonServiceAccount); err != nil {
 			log.Error(err, "failed to sync linked secrets for Component", "Component", component.Name, l.Action, l.ActionUpdate)
 			// do not break sync because of a faulty item
 		}
--- a/internal/controller/appstudio-pipeline-sa-watcher-controller.go
+++ b/internal/controller/appstudio-pipeline-sa-watcher-controller.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2025 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	appstudiov1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
+	l "github.com/konflux-ci/build-service/pkg/logs"
+)
+
+// TODO delete the controller after migration to the new dedicated to build Service Account.
+// AppstudioPipelineServiceAccountWatcherReconciler watches appstudio-pippeline Service Account
+// and syncs linked secrets updates to dedicated for build pipeline Service Account.
+type AppstudioPipelineServiceAccountWatcherReconciler struct {
+	Client client.Client
+	Scheme *runtime.Scheme
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *AppstudioPipelineServiceAccountWatcherReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.ServiceAccount{}, builder.WithPredicates(predicate.Funcs{
+			CreateFunc: func(e event.CreateEvent) bool {
+				return false
+			},
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				// React only if appstudio-pipeline Service Account is changed
+				sa, ok := e.ObjectNew.(*corev1.ServiceAccount)
+				if !ok {
+					return false
+				}
+				return sa.Name == buildPipelineServiceAccountName
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return false
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				return false
+			},
+		})).
+		Named("AppstudioPipelineServiceAccountWatcher").
+		Complete(r)
+}
+
+func (r *AppstudioPipelineServiceAccountWatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrllog.FromContext(ctx).WithName("appstudioPipelineSA")
+	log.Info("Synching secrets")
+
+	componentList := &appstudiov1alpha1.ComponentList{}
+	if err := r.Client.List(ctx, componentList, client.InNamespace(req.Namespace)); err != nil {
+		log.Info("failed to list Components")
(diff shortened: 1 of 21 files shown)


pkg/git/gitlab/gitlab_client.go
```diff
@@ -274,6 +274,11 @@ func (g *GitlabClient) GetBranchSha(repoUrl, branchName string) (string, error)
 	return sha, nil
 }
 
+func (g *GitlabClient) DownloadFileContent(repoUrl, branchName, filePath string) ([]byte, error) {
+	// TODO
```
mmorhun, line 278: To be implemented in this PR

internal/controller/appstudio-pipeline-sa-watcher-controller.go
```diff
@@ -0,0 +1,102 @@
...
+func (r *AppstudioPipelineServiceAccountWatcherReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := ctrllog.FromContext(ctx).WithName("appstudioPipelineSA")
+	log.Info("Synching secrets")
+
+	componentList := &appstudiov1alpha1.ComponentList{}
+	if err := r.Client.List(ctx, componentList, client.InNamespace(req.Namespace)); err != nil {
+		log.Info("failed to list Components")
+		return ctrl.Result{}, err
+	}
+
```
rcerven, line 81: you should get appstudio-pipeline SA here

internal/controller/appstudio-pipeline-sa-watcher-controller.go
```diff
@@ -0,0 +1,102 @@
...
+			if !errors.IsNotFound(err) {
+				log.Error(err, fmt.Sprintf("failed to read build pipeline Service Account %s in namespace %s", buildPipelineServiceAccountName, component.Namespace), l.Action, l.ActionView)
+				// do not break sync because of a faulty item
+			}
+			// Dedicated build pipeline Service Account hasn't been yet created.
+			// Skip for now, the sync will be performed on the Service Account creation.
+			continue
+		}
+
+		if err := LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx, r.Client, &component); err != nil {
```
rcerven, line 95: pass also compoent & pipeline SA, so you don't have to query it inside again

internal/controller/component_build_controller_service_account.go
```diff
@@ -0,0 +1,523 @@
...
+	if err := r.setServiceAccountInPipelineDefinition(ctx, component); err != nil {
+		log.Error(err, "failed to create migration PR")
+		return err
+	}
+
+	return nil
+}
+
+func (r *ComponentBuildReconciler) linkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx context.Context, component *appstudiov1alpha1.Component) error {
+	return LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx, r.Client, component)
```
rcerven, line 314: I am guessing this is because of tests

internal/controller/component_build_controller_service_account.go
```diff
@@ -0,0 +1,523 @@
...
+	return nil
+}
+
+// linkCommonSecretsToBuildPipelineServiceAccount searches for common build Secrets in Component namespace
+// marked with corresponding label and adds them into secrets section of the given Service Account.
+func (r *ComponentBuildReconciler) linkCommonSecretsToBuildPipelineServiceAccount(ctx context.Context, serviceAccount *corev1.ServiceAccount) error {
+	// Get all secrets with common for all Component label
+	commonSecretsList := &corev1.SecretList{}
+
+	commonBuildSecretRequirement, err := labels.NewRequirement(commonBuildSecretLabelName, selection.Equals, []string{"true"})
```
rcerven, line 110: what is setting that annotation?

internal/controller/component_build_controller_pac.go
```diff
@@ -423,6 +423,15 @@ func generateMergeRequestSourceBranch(component *appstudiov1alpha1.Component) st
 	return fmt.Sprintf("%s%s", pacMergeRequestSourceBranchPrefix, component.Name)
 }
 
+// getPipelineRunDefinitionFilePath returns full path in git repository to the pipeline run definition of the given Component.
+func getPipelineRunDefinitionFilePath(component *appstudiov1alpha1.Component, isPullRequest bool) string {
```
rcerven, line 427: probably better in internal/controller/component_build_controller_pipeline.go

rcerven: just FYI this will work ONLY if they didn't rename pipeline run files (about which we can't do much)

internal/controller/component_build_controller_service_account.go
```diff
@@ -0,0 +1,523 @@
...
+}
+
+// addServiceAccountToPipelineRun specifies given service account to use in the pipeline.
+func addServiceAccountToPipelineRun(pipelineRunBytes []byte, serviceAccountName string) ([]byte, error) {
+	buildPipelineData := &tektonapi.PipelineRun{}
+	if err := yaml.Unmarshal(pipelineRunBytes, buildPipelineData); err != nil {
+		return nil, err
+	}
+
+	buildPipelineData.Spec.TaskRunTemplate.ServiceAccountName = serviceAccountName
```
rcerven, line 520: you should 1st check if it isn't there already

and return additional boolean value indicating that

internal/controller/component_build_controller_service_account.go
```diff
@@ -0,0 +1,523 @@
...
+		}
+	}
+	if pullPipelineDefinitionFound {
+		pullPipelineContent, err = addServiceAccountToPipelineRun(pullPipelineContent, pipelineRunServiceAccountName)
+		if err != nil {
+			log.Error(err, "failed to add Service Account to pull pipeline definition")
+			return err
+		}
+	}
+
```
rcerven, line 483: addServiceAccountToPipelineRun should return also boolean if SA was added or not AND
only after that if it needs to be modified create MR

internal/controller/component_build_controller_pipeline.go
```diff
@@ -110,10 +110,26 @@ func (r *ComponentBuildReconciler) generatePaCPipelineRunConfigs(ctx context.Con
 		return nil, nil, err
 	}
 
+	// TODO delete the conditions after migration to the new Service Account done.
+	// If a Config Map with name "use-new-sa" exist, switch to the new Service Account.
+	useOldSA := true
+	if err := r.Client.Get(ctx, types.NamespacedName{Namespace: BuildServiceNamespaceName, Name: "use-new-sa"}, &corev1.ConfigMap{}); err == nil {
```
rcerven, line 116: I don't understand purpose of this CM

in reconcile you are ensuring that new SA exists, so why bother with this?

all new onboardings should use new one already

internal/controller/appstudio-pipeline-sa-watcher-controller.go
mmorhun: ok

internal/controller/appstudio-pipeline-sa-watcher-controller.go
mmorhun: ok

internal/controller/component_build_controller_service_account.go
mmorhun: According to the design doc, UI should set it along with linking the secret to all existing Components if user marks a secret as common for all existing and new Components.

internal/controller/component_build_controller_service_account.go
```diff
@@ -0,0 +1,523 @@
...
+	return nil
+}
+
+func (r *ComponentBuildReconciler) linkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx context.Context, component *appstudiov1alpha1.Component) error {
+	return LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx, r.Client, component)
+}
+
+// LinkCommonSecretsToNewServiceAccount links all secrets from appstudio-pipeline Service Account
+// except image repository secrets to the new dedicated to build Service Account.
+func LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx context.Context, c client.Client, component *appstudiov1alpha1.Component) error {
```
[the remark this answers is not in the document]
mmorhun: This function is also used when the migration is run.

internal/controller/component_build_controller_pac.go
mmorhun: > probably better in internal/controller/component_build_controller_pipeline.go

By name - yes, but in fact it's not used there, so I'd prefer to keep it closer to the main point of usage.

internal/controller/component_build_controller_service_account.go
mmorhun: Yeah, makes sense, than you, fixed.

internal/controller/component_build_controller_pipeline.go
mmorhun: This would allow us to perform migration when we are ready to do it, instead of migrating on operator deployment.

tkdchen (commented):
After migration to the new SA arch is done, both `internal/controller/component_build_controller_service_account.go` and the sync controller can be removed?

Can you describe every commit in the commit message as well? When going through the commit history, the git-log will be helpful and straightforward, whereas finding out the corresponding pull request requires more effort.

internal/controller/appstudio-pipeline-sa-watcher-controller.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		return ctrl.Result{}, err
+	}
+
+	for _, component := range componentList.Items {
+		dedicatedBuildPipelineServiceAccountName := getBuildPipelineServiceAccountName(&component)
+		buildPipelinesServiceAccount := &corev1.ServiceAccount{}
+		if err := r.Client.Get(ctx, types.NamespacedName{Name: dedicatedBuildPipelineServiceAccountName, Namespace: req.Namespace}, buildPipelinesServiceAccount); err != nil {
+			if !errors.IsNotFound(err) {
+				log.Error(err, fmt.Sprintf("failed to read build pipeline Service Account %s in namespace %s", dedicatedBuildPipelineServiceAccountName, component.Namespace), l.Action, l.ActionView)
```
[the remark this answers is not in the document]
rcerven: !errors.IsNotFound(err)

internal/controller/component_build_controller_service_account.go
```diff
@@ -0,0 +1,488 @@
...
+			}
+		}
+	}
+	return false
+}
+
+const saMigrationMergeRequestDescription = `
+## Build pipeline Service Account migration
+
+This PR chnages Service Account used by build pipeline from "appstudio-pipeline" to dedicated to the Component Service Account.
```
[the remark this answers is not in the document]
mmorhun: Thanks, fixed

internal/controller/appstudio-pipeline-sa-watcher-controller.go
```diff
@@ -0,0 +1,111 @@
...
+				// do not break sync because of a faulty item
+			}
+			// Dedicated build pipeline Service Account hasn't been yet created.
+			// Skip for now, the sync will be performed on the Service Account creation.
+			continue
+		}
+
+		if err := LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx, r.Client, &component, commonServiceAccount); err != nil {
+			log.Error(err, "failed to sync linked secrets for Component", "Component", component.Name, l.Action, l.ActionUpdate)
+			// do not break sync because of a faulty item
```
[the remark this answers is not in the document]
mmorhun: It should retry on the next sync

internal/controller/appstudio-pipeline-sa-watcher-controller.go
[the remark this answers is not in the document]
mmorhun: You've probably missed negation

internal/controller/component_build_controller_service_account.go
```diff
@@ -0,0 +1,488 @@
...
+// LinkCommonSecretsToNewServiceAccount links all secrets from appstudio-pipeline Service Account
+// except image repository secrets to the new dedicated to build Service Account.
+func LinkCommonAppstudioPipelineSecretsToNewServiceAccount(ctx context.Context, c client.Client, component *appstudiov1alpha1.Component) error {
+	log := ctrllog.FromContext(ctx)
+
+	appstudioPipelineServiceAccount := &corev1.ServiceAccount{}
+	if err := c.Get(ctx, types.NamespacedName{Name: buildPipelineServiceAccountName, Namespace: component.Namespace}, appstudioPipelineServiceAccount); err != nil {
+		if errors.IsNotFound(err) {
+			// Ignore not found error, assume the migration for the linked secrets has been done.
+			log.Info("skipping linked secret migration")
```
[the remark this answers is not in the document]
mmorhun: > Should be s/linked/linking/ because nothing is linked yet?

No, because we sync already linked secrets between Service Accounts.

Changed the message.

mmorhun:
@tkdchen 
> After migration to the new SA arch is done, both internal/controller/component_build_controller_service_account.go and the sync controller can be removed?

Only sync controller and migration part (see comment) of `component_build_controller_service_account.go` could be removed.

Regarding commits, I'll squash them with a description, because this should be one big change.

rcerven (approved):

tkdchen (approved):

PR #428, opened 2025-04-23
Do not sync dockercfg secret of appstudio-pipeline Service Account

mmorhun:
Do not sync dockercfg secret of `appstudio-pipeline` Service Account to dedicated build pipeline Service Accounts.
Clean the dockercfg secrets linked to dedicated build pipeline Service Accounts.

### DIFF
--- a/internal/controller/appstudio-pipeline-sa-watcher-controller_test.go
+++ b/internal/controller/appstudio-pipeline-sa-watcher-controller_test.go
@@ -159,6 +159,50 @@ var _ = Describe("appstudio-pipeline Service Account watcher controller", func()
 			}, timeout, interval).Should(BeTrue())
 		})
 
+		It("should ignore dockercfg secret of appstudio-pipeline", func() {
+			appstudioPipelineDockercfgSecretName := "appstudio-pipeline-dockercfg-c66gp"
+
+			appstudioPipelineSA := waitServiceAccount(appstudioPipelineSAKey)
+			appstudioPipelineSA.Secrets = append(appstudioPipelineSA.Secrets, corev1.ObjectReference{Name: appstudioPipelineDockercfgSecretName})
+			appstudioPipelineSA.ImagePullSecrets = append(appstudioPipelineSA.ImagePullSecrets, corev1.LocalObjectReference{Name: appstudioPipelineDockercfgSecretName})
+			Expect(k8sClient.Update(ctx, &appstudioPipelineSA)).To(Succeed())
+
+			Consistently(func() bool {
+				component1SA := waitServiceAccount(component1SAKey)
+				component2SA := waitServiceAccount(component2SAKey)
+				return !isSecretLinkedToServiceAccount(component1SA, appstudioPipelineDockercfgSecretName, false) &&
+					!isSecretLinkedToServiceAccount(component1SA, appstudioPipelineDockercfgSecretName, true) &&
+					!isSecretLinkedToServiceAccount(component2SA, appstudioPipelineDockercfgSecretName, false) &&
+					!isSecretLinkedToServiceAccount(component2SA, appstudioPipelineDockercfgSecretName, true)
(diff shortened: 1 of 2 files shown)


internal/controller/appstudio-pipeline-sa-watcher-controller_test.go
```diff
@@ -159,6 +159,50 @@ var _ = Describe("appstudio-pipeline Service Account watcher controller", func()
...
+		It("should ignore dockercfg secret of appstudio-pipeline", func() {
+			appstudioPipelineDockercfgSecretName := "appstudio-pipeline-dockercfg-c66gp"
+
+			appstudioPipelineSA := waitServiceAccount(appstudioPipelineSAKey)
+			appstudioPipelineSA.Secrets = append(appstudioPipelineSA.Secrets, corev1.ObjectReference{Name: appstudioPipelineDockercfgSecretName})
+			appstudioPipelineSA.ImagePullSecrets = append(appstudioPipelineSA.ImagePullSecrets, corev1.LocalObjectReference{Name: appstudioPipelineDockercfgSecretName})
+			Expect(k8sClient.Update(ctx, &appstudioPipelineSA)).To(Succeed())
+
+			Consistently(func() bool {
+				component1SA := waitServiceAccount(component2SAKey)
```
chmeliik, line 171:
```suggestion
				component1SA := waitServiceAccount(component1SAKey)
```

internal/controller/appstudio-pipeline-sa-watcher-controller_test.go
```diff
@@ -159,6 +159,50 @@ var _ = Describe("appstudio-pipeline Service Account watcher controller", func()
...
+				component2SA := waitServiceAccount(component2SAKey)
+				return !isSecretLinkedToServiceAccount(component1SA, appstudioPipelineDockercfgSecretName, false) &&
+					!isSecretLinkedToServiceAccount(component1SA, appstudioPipelineDockercfgSecretName, true) &&
+					!isSecretLinkedToServiceAccount(component2SA, appstudioPipelineDockercfgSecretName, false) &&
+					!isSecretLinkedToServiceAccount(component2SA, appstudioPipelineDockercfgSecretName, true)
```
chmeliik, line 176: nitpick: the tests may be simpler if the function checked both `Secrets` and `ImagePullSecrets`
```suggestion
				return !isSecretLinkedToServiceAccount(component1SA, appstudioPipelineDockercfgSecretName) &&
					!isSecretLinkedToServiceAccount(component2SA, appstudioPipelineDockercfgSecretName) &&
```

chmeliik (approved):
LGTM, small comments for the tests

To confirm my understanding: the reason why we do not want to sync that secret is that this secret is auto-generated and only relevant for the appstudio-pipeline SA?

internal/controller/appstudio-pipeline-sa-watcher-controller_test.go
mmorhun: Good catch, thanks

internal/controller/appstudio-pipeline-sa-watcher-controller_test.go
mmorhun: Not really, because we check linked and not linked cases. What if a secret is linked only to secrets but not pull secrets?

mmorhun:
> To confirm my understanding: the reason why we do not want to sync that secret is that this secret is auto-generated and only relevant for the appstudio-pipeline SA?

@chmeliik yes, exactly. Unfortunately, I've missed this fact in the original PR.

internal/controller/appstudio-pipeline-sa-watcher-controller_test.go
chmeliik: Ah, I missed that, I thought we were only checking the not linked cases

chmeliik (approved):

PR #429, opened 2025-04-24
Fix not found condition when creating Service Account migration PR

mmorhun:


mkosiarc (approved):

internal/controller/component_build_controller_service_account.go
```diff
@@ -476,7 +476,7 @@ func (r *ComponentBuildReconciler) setServiceAccountInPipelineDefinition(ctx con
 	var pushPipelineContent []byte
 	pushPipelineContent, err = gitClient.DownloadFileContent(repoUrl, baseBranch, pushPipelinePath)
 	if err != nil {
-		if !errors.IsNotFound(err) {
+		if !errors.IsNotFound(err) && !strings.Contains(err.Error(), "not found") {
```
chmeliik, line 479: For my information, why does `errors.IsNotFound` not work properly in this case? Why do we need to check the error message?

chmeliik: Ah, I see

PR #436, opened 2025-05-07
Fix appstudio-pipeline Service Account Secrets sync for newly created Components

mmorhun:
Handle the situation when both the Component and the Image Repository were created at the same time and Image Controller operator hasn't yet set the owner reference for the Image Repository. Even in such case, the push secret should not be treated as common and should not be synced to other Components build pipeline Service Account.

### DIFF
--- a/internal/controller/appstudio-pipeline-sa-watcher-controller.go
+++ b/internal/controller/appstudio-pipeline-sa-watcher-controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


internal/controller/appstudio-pipeline-sa-watcher-controller.go
```diff
@@ -96,6 +98,14 @@ func (r *AppstudioPipelineServiceAccountWatcherReconciler) Reconcile(ctx context
...
 		return ctrl.Result{}, err
 	}
+	// Wait all Image Repositories are provisioned
+	for _, imageRepository := range imageRepositoryList.Items {
+		if imageRepository.Status.State == "" {
+			// Wait Image Controller to finish
+			log.Info("Waiting for Image Repository provision", "ImageRepositoryName", imageRepository.Name)
+			return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+	}
```
chmeliik, line 108: For my understanding, why do we need to wait here? Do we depend on the owner references in a way that can't be worked around by using the labels? (Similarly to what we're adding to the `isComponentImageRepositorySecret` function)

mmorhun: The status needs to be checked first, because checking labels is sorta workaround.

chmeliik (approved):

PR #449, opened 2025-06-24
Automerge GitHub Actions updates

tnevrlka:
There is a lot of pull requests created for the repository that update
dependencies.
Reviewing all of them manually takes time and usually isn't much more
thorough than checking if the CI passes anyway.

Introduce a new package rule that automerges GH Actions.
GH Actions should be safe to update because the particular action
would fail in the PR

Note: `allowedVersions` is changed from an array based on the recommendation from Renovate config lint

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -8,9 +8,14 @@
       "matchPackageNames": [
         "registry.access.redhat.com/ubi9/go-toolset"
       ],
-      "allowedVersions": [
-        "<2"
-      ]
+      "allowedVersions": "<2"
+    },
+    {
+      "description": "Automerge GitHub Actions updates",
+      "matchManagers": [
+        "github-actions"
+      ],
+      "automerge": true
     }
   ]
 }


renovate.json
```diff
@@ -8,9 +8,15 @@
...
-      "allowedVersions": [
-        "<2"
-      ]
+      "allowedVersions": "<2"
+    },
+    {
+      "description": "Automerge GitHub Actions and Go dependencies updates",
+      "matchManagers": [
+        "github-actions",
+        "gomod"
```
eskultety, line 17: Merging actions automatically is one thing, because yes, as you correctly note, the chance the CI would immediately fail with a problematic change is high, but I'd be more careful about language dependencies or any other direct repo dependencies. It can be passing today, but you introduce a Go code change tomorrow and that might not pass the CI leaving you wondering how's that possible if it's e.g. passing locally. That's just the theory though, I don't have enough visibility into this repo, but automerging can slip in potentially problematic changes completely undetected making the debug surface much larger (that of course also assumes that noone ever forgets to consider the fact that there in fact IS some automerging in place). To me, loosing the visibility in exchange for a slightly decreased traffic from renovate PRs isn't simply worth it. What we did in Hermeto to decrease the auto PR noise was tweaking the schedule to a monthly cycle (because that was still a reasonable time frame for meaningful external deps to be introduced to the project) and so we don't need to waste capacity for reviews more than absolutely necessary.

tnevrlka: Would you mind describing your concern `It can be passing today, but you introduce a Go code change tomorrow and that might not pass the CI leaving you wondering how's that possible if it's e.g. passing locally.` more in depth?

I don't understand how that situation could happen. 

The two "bad" scenarios that come to my mind are:

**Scenario 1: Go dependency breaks something, CI catches it**
1. Everything works
2. Update of dependency `foo` from `1.0` to `2.0` removed a function we're using, for example
3. PR is created, its CI checks fail because the function was removed so it's caught by lint
4. Since the PR is failing, a manual review is required

**Scenario 2: Go dependency breaks something, CI doesn't catch it**
1. Everything works
2. Update of dependency `bar` from `1.0` to `2.0` involves a change that introduces a rare regression which is not caught by tests
3. PR is created, its CI checks pass, the PR is merged
4. The wrong behavior is uncovered, a bug gets filed
5. The assignee of the bug starts do debug and either:
    1. They'll do some research and find out the issue is a known problem with `bar 2.0`
    2. They do a git bisect and find out the bug started occurring after merging the bad update PR

I'm not sure how the manual review process could find the issue - either it's an obvious enough problem that the CI catches it, or it's so hidden that a manual review wouldn't find it out anyway and it would be reported as a bug afterwards.

What things are you doing during a manual review that would uncover the issue even in a situation the CI passes?

Sidenote: The monthly schedule idea sounds pretty good, I would be okay with that as a compromise

eskultety: > The motivation behind the change was that most of the PRs do not even have any, see https://github.com/konflux-ci/build-service/pull/408, https://github.com/konflux-ci/build-service/pull/409 and https://github.com/konflux-ci/build-service/pull/410 for example

I see. Well, I only shared my 2c :) . On a side note, who thought it's such a brilliant idea to update a library (or pretty much anything) by a digest instead of using versioning? Fair to say though this is documented to be an experimental library, but still, in general.

rcerven:
I am not sure if we want to automerge go deps, @mmorhun ?

mmorhun:
After some analyzing, I think we should disable this go deps updates at all.
The reason for it is that we target Openshift which is usually behind k8s and using the newest version could potentially lead to a wrong behavior. Also, all k8s/controller-runtime deps must be updated in sync. Moreover it could get more tricky with transitive dependencies especially if they are both: direct and transitive.
The only reason for us to update go deps is to get CVEs fixes. About those we are notified by Conforma and update the deps manually when it's needed.

mmorhun (approved):

renovate.json
```diff
@@ -8,9 +8,14 @@
       "matchPackageNames": [
         "registry.access.redhat.com/ubi9/go-toolset"
       ],
-      "allowedVersions": [
-        "<2"
-      ]
+      "allowedVersions": "<2"
+    },
+    {
+      "description": "Automerge GitHub Actions and Go dependencies updates",
```
tisutisu, line 14: nit pick: do we need to update the description after the review?

tnevrlka: Good catch! Yes, I missed that. Thanks!

PR #450, opened 2025-06-26
Remove ability to onboard a Component with appstudio-pipeline Service Account

mmorhun:
Currently Build Service has option to onboard a Component with deprecated appstudio-pipeline Service Account. It should not be a case any more since appstudio-pipeline SA is past due date. All new Components should use dedicated build pipeline Service Account.


tnevrlka (approved):
LGTM

PR #455, opened 2025-07-09
Enable ARM64 builds for build-service

brianwcook:
This PR enables ARM64 builds for the build-service component by:

- Adding build-platforms parameter with linux/x86_64 and linux/arm64 support
- Converting single buildah-oci-ta tasks to matrix buildah-remote-oci-ta tasks
- Updating build-image-index to collect from matrix build results
- Setting ALWAYS_BUILD_INDEX to 'true' to force index generation for multi-platform builds

Changes applied to both push and pull-request pipelines.

### DIFF
--- a/.tekton/build-service-pull-request.yaml
+++ b/.tekton/build-service-pull-request.yaml
@@ -107,6 +107,12 @@ spec:
       description: Path to a file with build arguments for buildah, see https://www.mankier.com/1/buildah-build#--build-arg-file
       name: build-args-file
       type: string
+    - default:
+      - linux/x86_64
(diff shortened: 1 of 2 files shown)


brianwcook:
@mmorhun I don't have access to build workspace to see logs; can you tell me why its failing. I didn't change anything that should cause an e2e test failure that didn't already exist.

flacatus:
a good fix for this is to get rid off using openshift rosa for tests and use kind and Konflux upstream to deploy https://github.com/konflux-ci/build-service/pull/448

@mmorhun @tisutisu @chmeliik

PR #466, opened 2025-08-01
fix: when nudging, have Renovate be more patient with GitLab

compi-migui:
GitLab has a long-standing issue where the API refuses attempts to set "merge when pipeline succeeds" even when the MR is ready for it[1].

Renovate already does some retrying with exponential backoff[2], but the default behavior is to only wait for up to 7.5 seconds. This leads to nudging MRs very often failing to automerge, especially at times of high GitLab load.

By setting a higher base delay value and a higher number of retries we can ensure automerge works a lot more reliably.

[KONFLUX-9373](https://issues.redhat.com//browse/KONFLUX-9373)

[1] https://gitlab.com/gitlab-org/gitlab/-/issues/353984
[2] https://github.com/renovatebot/renovate/blob/88768303ee7f246da3337c5d34cace8c53f1284e/lib/modules/platform/gitlab/index.ts#L694-L714

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/internal/controller/renovate_util.go
+++ b/internal/controller/renovate_util.go
@@ -633,7 +633,7 @@ func (u ComponentDependenciesUpdater) CreateRenovaterPipeline(ctx context.Contex
 		// we are passing host rules via variable, because we can't resolve variable in json config
 		// also this way we can use custom provided config without any modifications
 		renovateCmds = append(renovateCmds,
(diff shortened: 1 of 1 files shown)


mmorhun:
@compi-migui thank you for the contribution!
The only thing I doubt is that we are setting the delay right. According to the [renovate docs](https://docs.renovatebot.com/self-hosted-experimental/#renovate_x_gitlab_auto_mergeable_check_attemps), it would take `RENOVATE_X_GITLAB_MERGE_REQUEST_DELAY * attempt * attempt` milliseconds. So, the last, 7-th attempt will take almost 6 minutes. Do we really need to wait that long?

mmorhun:
Tests failed due to infra issue:
```
konflux-kite-in-cluster-local                      Synced   Degraded
```

PR #478, opened 2025-09-04
fix(sealights): add missing params to e2e tests

MartinBasti:
Missing params in e2e test tasks causes missing coverage in sealights

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/integration-tests/pipelines/konflux-e2e-tests.yaml
+++ b/integration-tests/pipelines/konflux-e2e-tests.yaml
@@ -44,6 +44,9 @@ spec:
     - name: enable-sl-plugin
       description: "A flag to enable or disable the Sealights integration feature. When set to 'true', test results are sent to Sealights for analysis; otherwise, this feature is skipped."
       default: "true"
+    - name: enable-sealights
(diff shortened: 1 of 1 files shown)


integration-tests/pipelines/konflux-e2e-tests.yaml
```diff
@@ -44,6 +44,9 @@ spec:
     - name: enable-sl-plugin
```
mkosiarc, line 44: are both needed? Isn't this a duplicate? the description make it look like the same flag

MartinBasti: https://github.com/konflux-ci/e2e-tests/blob/7d287c1c637013e774a24364adbb1cb648285152/integration-tests/tasks/konflux-e2e-tests-task.yaml#L201  it's needed, otherwise step is skipped

the enable-sl-plugin is exported only as env var, dunno where it's used, it can be used by those go scripts, it's related to ginkgo somehow

PR #482, opened 2025-09-09
new annotation on nudged component "build.appstudio.openshift.io/buil…

rcerven:
…d-nudge-simple-branch"

when annotation is set with "true" nudged component branch won't have nudged component name in it

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually


lhh:
This is a good solution which preserves the older behavior! Thanks!

PR #486, opened 2025-09-29
feat: add required tagging of test clusters

psturc:
### Note: CI does not have to pass to merge this PR

Validated in AWS:

Screenshot 2025-10-01 at 10 42 09

### DIFF
--- a/integration-tests/pipelines/konflux-e2e-tests.yaml
+++ b/integration-tests/pipelines/konflux-e2e-tests.yaml
@@ -101,9 +101,7 @@ spec:
         - name: id
           value: $(context.pipelineRun.name)
         - name: tags
-          value: >
-            'owner=konflux-devprod@redhat.com,project=Konflux,created-by=integration-pipeline,
-            iac=mapt,cluster-name=$(context.pipelineRun.name),k8s-type=kind,component=build-service'
+          value: 'owner=konflux-devprod@redhat.com,project=Konflux,created-by=integration-pipeline,component=build-service'
         - name: debug
           value: 'false'
         - name: ownerKind


psturc:
> I'm really not sure if such things belong to the upstream repository

Thanks for feedback. I initiated a discussion about that in our team and a new set of tags was determined, omitting the internal details and optional tags that are not required, just optional.

The new proposed set of tags is necessary for tracking the usage of cloud resources with our tools.

cc @Dannyb48

integration-tests/pipelines/konflux-e2e-tests.yaml
```diff
@@ -101,7 +101,9 @@ spec:
         - name: id
           value: $(context.pipelineRun.name)
         - name: tags
-          value: env=konflux,user=build-service
+          value: >
+            'owner=konflux-devprod@redhat.com,project=Konflux,created-by=integration-pipeline,
+            iac=mapt,cluster-name=$(context.pipelineRun.name),k8s-type=kind,component=build-service'
```
chmeliik, line 106: Is it expected that we're setting cluster-name to the pipelineRun name?

What does `iac=mapt` mean?

chmeliik: I see :+1: 

The cluster name and the IaC tool seems like an implementation detail of the `provision-kind-cluster` task. Would it make more sense to add those tags in the task itself? That way, if the implementation of the task changes, you won't have to go and update the tags for all the users of the task

psturc: good point! I will update that Task

chmeliik (approved):

tisutisu (approved):

psturc:
can we force merge this PR?

PR #493, opened 2025-10-23
Remove SeaLights from build-service

tnevrlka:
SeaLights was introduced in order to provide E2E tests code coverage. However, the license for it has not been renewed and SeaLights will thus stop working.

Get rid of it before it breaks CI


chmeliik (approved):
A truly heartbreaking loss o7

PR #510, opened 2025-11-12
PoC: Preserve existing settings for PaC repositories

mh21:
Added functionality to preserve existing repository settings when updating a PaC repository. This ensures user-managed settings are not overwritten.

The Hummingbird Project uses a mono repository for its container images at https://gitlab.com/redhat/hummingbird/containers. Because the build controller spams the MRs until no human contribution is possible anymore, we have had our Repository object modified with

    {"spec":{"settings":{"gitlab":{"comment_strategy":"disable_all"}}}}

Turns out that this setting gets reverted whenever a new Component is onboarded to PaC.

Steps to reproduce:

- have a Component and associated Repository nicely set up
- disable comments as above
- create a new Component for the same git repo, but with `build.appstudio.openshift.io/status={"pac":{"state":"enabled"},"message":"done"}`
- check that the comment_strategy setting is still present, ie it is not the Component creation that triggers the revert
- annotate the new Component with `build.appstudio.openshift.io/request=configure-pac-no-mr`
- check that now the comment_strategy setting is reverted

The code in this PR was created by Copilot, when tasked with fixing the symptoms, so it might make no sense at all.

I also don't fully understand how the current code with Update() is actually loosing the settings object, but using Patch() conceptually seems to be the right thing here anyway.

/cc @scoheb

### DIFF
--- a/internal/controller/component_build_controller_pac_repository.go
+++ b/internal/controller/component_build_controller_pac_repository.go
@@ -74,12 +74,15 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
 	}
 	if repository != nil {
 		pacRepositoryOwnersNumber := len(repository.OwnerReferences)
+		// preserve original for patching
+		origRepo := repository.DeepCopy()
 		if err := controllerutil.SetOwnerReference(component, repository, r.Scheme); err != nil {
 			log.Error(err, "failed to add owner reference to existing PaC repository", "PaCRepositoryName", repository.Name)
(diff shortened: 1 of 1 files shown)


mmorhun:
Thank you for trying to fix the problem @mh21 .
This happens because of older Repository structure used in Build Service.
I've created update [PR here](https://github.com/konflux-ci/build-service/pull/511).

rcerven:
closing in favor of https://github.com/konflux-ci/build-service/pull/511

PR #526, opened 2025-12-01
Disable renovate updates for golang dependencies

mmorhun:
We have a lot of PRs from renovate that bumps some golang deps.
It causes noise for no gain:
- Security related issues are reported by Konflux CI
- Quite often upating a depencency involves other updates, since some deps must be in sync.
- If everything just works, why would we need a newer version? I'd prefer stability.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -18,10 +18,22 @@
       "automerge": true
     },
     {
+      "description": "Go minor updates",
       "matchManagers": ["gomod"],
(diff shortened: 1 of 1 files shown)


brunoapimentel (approved):
LGTM, but I'd drop the https://github.com/konflux-ci/build-service/pull/526/commits/df4da06f3cadf43c43ff5a07fca4a84b9d6e7b17 commit.

PR #527, opened 2025-12-02
test: collect coverage even if tests fail

psturc:
If we collect the e2e test coverage data **only when all e2e tests pass**, we're losing data for jobs that are failing due to flakey tests (e.g. periodic jobs)

Adding the "collect-and-upload-coverage" task to the finally block should prevent losing the coverage data.

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

### DIFF
--- a/integration-tests/pipelines/konflux-e2e-tests.yaml
+++ b/integration-tests/pipelines/konflux-e2e-tests.yaml
@@ -163,9 +163,8 @@ spec:
           value: kfg-$(context.pipelineRun.name)
         - name: test-environment
           value: "upstream"
+  finally:
     - name: collect-and-upload-coverage
-      runAfter:
-        - konflux-e2e-tests
       params:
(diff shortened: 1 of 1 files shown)


integration-tests/pipelines/konflux-e2e-tests.yaml
```diff
@@ -174,7 +173,7 @@ spec:
         - name: test-name
           value: e2e-tests
         - name: oci-container
-          value: "$(params.oci-container-repo):$(context.pipelineRun.name)"
+          value: "$(params.oci-container-repo):$(context.pipelineRun.name)-coverport"
```
psturc: Keeping the coverage data in a separate OCI artifact to avoid an issue when data is pushed to the same tag concurrently (from 2 tasks in finally block - already seen in CI)

PR #583, opened 2026-04-02
Implement skip-builds options fot new component model

rcerven:
[STONEBLD-4532](https://redhat.atlassian.net/browse/STONEBLD-4532)

### Checklist:
 - Write / update unit tests
 - Ensure `make test` passes
 - Test code changes manually

[STONEBLD-4532]: https://redhat.atlassian.net/browse/STONEBLD-4532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/controller/component_build_controller_pac_repository.go
+++ b/internal/controller/component_build_controller_pac_repository.go
@@ -23,6 +23,7 @@ import (
 	"reflect"
 	"regexp"
 	"slices"
+	"strconv"
 	"strings"
 
 	compapiv1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
@@ -38,6 +39,11 @@ import (
 	"github.com/konflux-ci/build-service/pkg/boerrors"
 	"github.com/konflux-ci/build-service/pkg/common"
 	l "github.com/konflux-ci/build-service/pkg/logs"
+)
+
+const (
+	BuildsEnabledParamName   = "builds_enabled"
+	PacTargetBranchFilterKey = "pac.target_branch"
 )
 
 // TODO remove newModel handling after only new model is used
@@ -89,8 +95,61 @@ func getUniqueStrings(items []string) []string {
 	return result
 }
 
+// updatePaCRepositorySkipBuildsParams updates the PaC repository parameters based on the skip builds map.
+// Returns true if any parameters were updated, false otherwise.
+func updatePaCRepositorySkipBuildsParams(repository *pacv1alpha1.Repository, skipBuildsMap map[string]bool) bool {
+	// Build desired params based on skipBuildsMap
+	desiredParams := []pacv1alpha1.Params{}
+	for revision, skipBuilds := range skipBuildsMap {
+		// The applyParamFilter defines when the parameter is passed to the pipeline run.
+		applyParamFilter := getPaCParamFilterForRevision(revision)
+		desiredParams = append(desiredParams, pacv1alpha1.Params{
+			Name:   BuildsEnabledParamName,
+			Value:  strconv.FormatBool(!skipBuilds),
+			Filter: applyParamFilter,
+		})
+	}
+
+	// Check if params need to be updated
+	if repository.Spec.Params == nil {
+		repository.Spec.Params = &desiredParams
+		return true
+	}
+
+	updatedParams := *repository.Spec.Params
+	paramsUpdated := false
+
+	// Check each desired param against existing params
+	// Params for removed branches are removed during version cleanup
+	for _, desired := range desiredParams {
+		found := false
+		for i := range updatedParams {
+			if updatedParams[i].Name == desired.Name && updatedParams[i].Filter == desired.Filter {
+				found = true
+				// Update value if different
+				if updatedParams[i].Value != desired.Value {
+					updatedParams[i].Value = desired.Value
+					paramsUpdated = true
+				}
+				break
+			}
+		}
+		// Add new param if not found
+		if !found {
+			updatedParams = append(updatedParams, desired)
+			paramsUpdated = true
+		}
+	}
+
+	if paramsUpdated {
+		repository.Spec.Params = &updatedParams
+	}
+
+	return paramsUpdated
+}
+
 // TODO remove newModel handling after only new model is used
-func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, component *compapiv1alpha1.Component, pacConfig *corev1.Secret, repositorySettings *compapiv1alpha1.RepositorySettings, newModel bool) (string, error) {
+func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, component *compapiv1alpha1.Component, pacConfig *corev1.Secret, repositorySettings *compapiv1alpha1.RepositorySettings, skipBuildsMap map[string]bool, newModel bool) (string, error) {
 	log := ctrllog.FromContext(ctx)
 
 	// Check multi component git repository scenario.
@@ -111,6 +170,7 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
 		}
 
 		repositorySettingsUpdated := false
+		repositoryParamsUpdated := false
 		if newModel {
 			// Initialize Settings if nil
 			if repository.Spec.Settings == nil {
@@ -149,9 +209,13 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
 					repositorySettingsUpdated = true
 				}
 			}
+
+			if skipBuildsMap != nil {
+				repositoryParamsUpdated = updatePaCRepositorySkipBuildsParams(repository, skipBuildsMap)
+			}
 		}
 
-		if (len(repository.OwnerReferences) > pacRepositoryOwnersNumber) || repositorySettingsUpdated {
+		if (len(repository.OwnerReferences) > pacRepositoryOwnersNumber) || repositorySettingsUpdated || repositoryParamsUpdated {
 			if err := r.Client.Update(ctx, repository); err != nil {
 				log.Error(err, "failed to update existing PaC repository with component owner reference", "PaCRepositoryName", repository.Name)
 				return "", err
@@ -164,7 +228,7 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
 	}
 
 	// This is the first Component that does PaC provision for the git repository
-	repository, err = generatePACRepository(*component, pacConfig, repositorySettings, newModel)
+	repository, err = generatePACRepository(*component, pacConfig, repositorySettings, skipBuildsMap, newModel)
 	if err != nil {
 		return "", err
 	}
@@ -265,7 +329,7 @@ func pacRepoAddParamWorkspaceName(repository *pacv1alpha1.Repository, workspaceN
 
 // TODO remove newModel handling after only new model is used
 // generatePACRepository creates configuration of Pipelines as Code repository object.
-func generatePACRepository(component compapiv1alpha1.Component, config *corev1.Secret, repositorySettings *compapiv1alpha1.RepositorySettings, newModel bool) (*pacv1alpha1.Repository, error) {
+func generatePACRepository(component compapiv1alpha1.Component, config *corev1.Secret, repositorySettings *compapiv1alpha1.RepositorySettings, skipBuildsMap map[string]bool, newModel bool) (*pacv1alpha1.Repository, error) {
 	gitProvider, err := getGitProvider(component, newModel)
 	if err != nil {
 		return nil, err
@@ -363,12 +427,21 @@ func generatePACRepository(component compapiv1alpha1.Component, config *corev1.S
 		}
 	}
 
+	if newModel && skipBuildsMap != nil {
+		params := []pacv1alpha1.Params{}
+		for revision, skipBuild := range skipBuildsMap {
+			applyParamFilter := getPaCParamFilterForRevision(revision)
+			params = append(params, pacv1alpha1.Params{Name: BuildsEnabledParamName, Value: strconv.FormatBool(!skipBuild), Filter: applyParamFilter})
+		}
+		repository.Spec.Params = &params
+	}
+
 	return repository, nil
 }
 
 // TODO remove newModel handling after only new model is used
-// cleanupPaCRepositoryIncomings cleans up incomings in Repository
-func (r *ComponentBuildReconciler) cleanupPaCRepositoryIncomings(ctx context.Context, component *compapiv1alpha1.Component) error {
+// cleanupPaCRepositoryIncomingsAndSkipBuildParams cleans up incomings and skip build params in Repository
+func (r *ComponentBuildReconciler) cleanupPaCRepositoryIncomingsAndSkipBuildParams(ctx context.Context, component *compapiv1alpha1.Component) error {
 	log := ctrllog.FromContext(ctx)
 	newModel := true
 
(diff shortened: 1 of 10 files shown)


internal/controller/component_build_controller_pac_repository.go
```diff
@@ -40,6 +41,11 @@ import (
 	l "github.com/konflux-ci/build-service/pkg/logs"
 )
 
+const (
+	BuildsEnabledParamName = "building_enabled"
```
mmorhun: `builds_enabled`

internal/controller/component_build_controller_pac_repository.go
```diff
@@ -149,9 +156,55 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
...
+				for revision, skipBuilds := range skipBuildsMap {
+					filter := fmt.Sprintf("%s == %s", PacTargetBranchFilter, revision)
+					desiredParams = append(desiredParams, pacv1alpha1.Params{
+						Name:   BuildsEnabledParamName,
+						Value:  strconv.FormatBool(!skipBuilds),
+						Filter: filter,
+					})
+				}
+
+				// Check if params need to be updated
```
mmorhun: I think it could be a separate function

internal/controller/component_build_controller_pac_repository.go
```diff
@@ -149,9 +156,55 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
...
+				if repository.Spec.Params == nil {
+					repository.Spec.Params = &desiredParams
+					repositoryParamsUpdated = true
+				} else {
+					newParams := *repository.Spec.Params
```
mmorhun: nitpick: they are not really new, they are more updated params.

internal/controller/component_build_controller_pac_repository.go
```diff
@@ -40,6 +41,11 @@ import (
 	l "github.com/konflux-ci/build-service/pkg/logs"
 )
 
+const (
+	BuildsEnabledParamName = "building_enabled"
+	PacTargetBranchFilter  = "pac.target_branch"
```
mmorhun: It's not really a filter, but name used in filter, we need to find a better name, maybe `PacTargetBranchFilterKey`?

internal/controller/component_build_controller_pac_repository.go
```diff
@@ -149,9 +156,55 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
 					repositorySettingsUpdated = true
 				}
 			}
+
+			if skipBuildsMap != nil {
+				// Build desired params based on skipBuildsMap
+				desiredParams := []pacv1alpha1.Params{}
+				for revision, skipBuilds := range skipBuildsMap {
+					filter := fmt.Sprintf("%s == %s", PacTargetBranchFilter, revision)
```
mmorhun:
```suggestion
					// The filter defines when the parameter is passed to the pipeline run.
					applyParamFilter := fmt.Sprintf("%s == %s", PacTargetBranchFilter, revision)
```

internal/controller/component_build_controller_pac_repository.go
```diff
@@ -149,9 +156,55 @@ func (r *ComponentBuildReconciler) ensurePaCRepository(ctx context.Context, comp
...
+								}
+								break
+							}
+						}
+						// Add new param if not found
+						if !found {
+							newParams = append(newParams, desired)
+							repositoryParamsUpdated = true
+						}
+					}
```
mmorhun: Would be nice to leave a comment about removal (that it is handled in another place).

internal/controller/component_build_controller_pac_repository.go
```diff
@@ -363,12 +416,21 @@ func generatePACRepository(component compapiv1alpha1.Component, config *corev1.S
 		}
 	}
 
+	if newModel && skipBuildsMap != nil {
+		params := []pacv1alpha1.Params{}
+		for revision, skipBuild := range skipBuildsMap {
+			filter := fmt.Sprintf("%s == %s", PacTargetBranchFilter, revision)
```
mmorhun: Do you think `filter := getPaCParamFilterForRevision(revision)` makes sense?

internal/controller/component_build_controller_pipeline.go
```diff
@@ -770,6 +770,8 @@ func generateCelExpressionForPipeline(component *compapiv1alpha1.Component, gitC
 	eventCondition := fmt.Sprintf(`event == "%s"`, eventType)
 
 	targetBranchCondition := fmt.Sprintf(`target_branch == "%s"`, versionInfo.Revision)
+	buildingEnabledCondition := fmt.Sprintf(`%s == "true"`, BuildsEnabledParamName)
```
mmorhun: `buildingEnabledCondition` -> `buildEnabledCondition`

internal/controller/component_build_controller_pipeline.go
```diff
@@ -806,7 +808,7 @@ func generateCelExpressionForPipeline(component *compapiv1alpha1.Component, gitC
 		pathChangedSuffix = fmt.Sprintf(` && ( "%s***".pathChanged() || ".tekton/%s".pathChanged() %s)`, contextDir, pipelineFileName, dockerfilePathChangedSuffix)
 	}
 
-	return fmt.Sprintf("%s && %s%s", eventCondition, targetBranchCondition, pathChangedSuffix), nil
+	return fmt.Sprintf("%s && %s && %s%s", eventCondition, targetBranchCondition, buildingEnabledCondition, pathChangedSuffix), nil
```
mmorhun: Can we disable build only when the param is false, so unset means build enabled?

internal/controller/component_build_controller_v2.go
```diff
@@ -201,6 +203,9 @@ func (r *ComponentBuildReconciler) ReconcileNewModel(ctx context.Context, req ct
 		// Repository exists but settings changed, need to update
 		createOrUpdateRepository = true
 		log.Info("Repository settings changed, need to update PAC Repository", l.Action, l.ActionUpdate, "PacRepository", component.Status.PacRepository, "SpecRepositorySettings", component.Spec.RepositorySettings, "StatusRepositorySettings", component.Status.RepositorySettings)
+	} else if skipBuildsChanged {
+		createOrUpdateRepository = true
+		log.Info("Skip builds settings changed, need to update PAC Repository", l.Action, l.ActionUpdate, "PacRepository", component.Status.PacRepository)
```
mmorhun, line 209: nitpick: the message will not be printed if both repository settings and params changed.

internal/controller/component_build_controller_v2.go
```diff
@@ -1287,6 +1357,39 @@ func (r *ComponentBuildReconciler) removeVersionsFromStatus(ctx context.Context,
...
+// updateVersionsSkipBuilds updates the SkipBuilds field for versions in component.Status.Versions
+// based on the values in existingSpecVersions. Only updates status if there are actual changes.
+// Returns an error if the status update fails.
+func (r *ComponentBuildReconciler) updateVersionsSkipBuilds(ctx context.Context, component *compapiv1alpha1.Component, existingSpecVersions map[string]*VersionInfo) error {
+	log := ctrllog.FromContext(ctx)
+
+	hasChanges := false
+
+	// Loop through status versions and update SkipBuilds from spec if different
+	for i := range component.Status.Versions {
```
mmorhun, line 1370: I think we should loop through spec and compare with status.
Also, should we ignore failed versions?

internal/controller/component_build_controller_v2.go
```diff
@@ -1030,8 +1099,9 @@ func (r *ComponentBuildReconciler) updateVersionsStatus(ctx context.Context, com
 	for _, version := range versions {
 		versionInfo := existingSpecVersions[version.name]
 		versionStatus := compapiv1alpha1.ComponentVersionStatus{
-			Name:     version.name,
-			Revision: versionInfo.Revision,
+			Name:       version.name,
+			Revision:   versionInfo.Revision,
+			SkipBuilds: versionInfo.SkipBuilds,
```
mmorhun, line 1105: In most cases, SkipBuilds will be omitted in spec versions (don't skip), but always preset in status.
Is it desired behavior? Should we set the field in status explicitly if it is set explicitly in spec? I'm not sure if it would worth the added complexity.

internal/controller/component_build_controller_v2.go
```diff
@@ -763,6 +776,62 @@ func buildVersionInfoMap(component *compapiv1alpha1.Component, fromStatus bool)
...
 }
 
+// buildSkipBuildsMapAndCheckChanges creates a map of revisions to their SkipBuilds values from spec
+// and checks if any version has a different SkipBuilds value in status or is missing from status.
+// For multiple versions with the same revision: SkipBuilds is false if ANY version has SkipBuilds=false,
+// and true ONLY if ALL versions for that revision have SkipBuilds=true.
+// Returns:
+//   - skipBuildsMap: map of revision to their SkipBuilds values (aggregated by revision)
+//   - hasChanges: true if any version is missing from status or has different SkipBuilds value in status
+func buildSkipBuildsMapAndCheckChanges(existingSpecVersions map[string]*VersionInfo, component *compapiv1alpha1.Component) (map[string]bool, bool) {
```
mmorhun: Usually, component is the first parameter (unless ctx is passed).

internal/controller/component_build_controller_v2.go
```diff
@@ -763,6 +776,62 @@ func buildVersionInfoMap(component *compapiv1alpha1.Component, fromStatus bool)
...
+	// This ensures versions will use the correct aggregated value when updating status,
+	// since multiple versions can share the same revision with different SkipBuilds settings in spec.
+	for _, versionInfo := range existingSpecVersions {
+		versionInfo.SkipBuilds = skipBuildsMap[versionInfo.Revision]
+	}
+
+	// Check if any version has different SkipBuilds in status or is missing from status
+	for versionName, versionInfo := range existingSpecVersions {
+		// Check if version exists in status
+		found := false
```
mmorhun: nitpick: `found` -> `foundInStatus`

internal/controller/component_build_controller_v2.go
```diff
@@ -763,6 +776,62 @@ func buildVersionInfoMap(component *compapiv1alpha1.Component, fromStatus bool)
...
+		for _, statusVersion := range component.Status.Versions {
+			if statusVersion.Name == versionName {
+				found = true
+				// Check if SkipBuilds value is different
+				if statusVersion.SkipBuilds != versionInfo.SkipBuilds {
+					skipBuildsChanged = true
+					break
```
mmorhun: not needed

internal/controller/component_build_controller_v2_test.go
```diff
@@ -1054,12 +1109,163 @@ var _ = Describe("Component build controller new model", func() {
...
+				if len(*repository.Spec.Params) != 1 {
+					return false
+				}
+				return (*repository.Spec.Params)[0].Value == "false"
+			}, timeout, interval).Should(BeTrue())
+
+			skipBuildsMap = map[string]bool{"main": true}
+			// Verify Repository CR was created with correct URL and no GitProvider (GitHub App)
+			validatePaCRepository(component, "", "", skipBuildsMap)
+		})
```
mmorhun, line 1203: Should we test restoring build here too?

internal/controller/component_build_controller_v2.go
```diff
@@ -763,6 +776,62 @@ func buildVersionInfoMap(component *compapiv1alpha1.Component, fromStatus bool)
 	return versionInfoMap
 }
 
+// buildSkipBuildsMapAndCheckChanges creates a map of revisions to their SkipBuilds values from spec
+// and checks if any version has a different SkipBuilds value in status or is missing from status.
+// For multiple versions with the same revision: SkipBuilds is false if ANY version has SkipBuilds=false,
+// and true ONLY if ALL versions for that revision have SkipBuilds=true.
```
mmorhun, line 784: Should we set a message in the version status if the version has `SkipBuilds` set to `true` in spec, but `false` in status? Otherwise the users will be confused.

internal/controller/component_build_controller_pac_repository.go
rcerven: ok

internal/controller/component_build_controller_pac_repository.go
rcerven: ok

internal/controller/component_build_controller_pac_repository.go
rcerven: ok

internal/controller/component_build_controller_pac_repository.go
rcerven: ok

internal/controller/component_build_controller_pac_repository.go
rcerven: yeah, created

internal/controller/component_build_controller_pipeline.go
rcerven: ok

internal/controller/component_build_controller_pipeline.go
rcerven: yes on-cel has to compare param value, if that param is removed from on-cel build will be enabled

internal/controller/component_build_controller_pac_repository.go
rcerven: ok

internal/controller/component_build_controller_v2.go
rcerven: moved to separate if

internal/controller/component_build_controller_v2.go
rcerven: this method is called to react to ONLY skip builds changes

existingSpecVersions  has versions from spec

so all we want from this method to do is update those in status which won't be modified with onboarding method

we are updating Repository, it doesn't matter if versions are failed

internal/controller/component_build_controller_pac_repository.go
rcerven: moved to a function

internal/controller/component_build_controller_v2.go
rcerven: yes that is desired behavior, so status is explicit everywhere about it

internal/controller/component_build_controller_v2.go
rcerven: changed

internal/controller/component_build_controller_v2.go
rcerven: changed

internal/controller/component_build_controller_v2.go
rcerven: removed

internal/controller/component_build_controller_v2.go
rcerven: message is used for errors

we will have to document it

internal/controller/component_build_controller_v2_test.go
rcerven: sure, added

internal/controller/component_build_controller_v2.go
rcerven: it will be confusing anyway if 2 components using the same repo & revisions will have different values

and Repository will be set based on last component changed

internal/controller/component_build_controller_pac_repository.go
```diff
@@ -415,19 +488,50 @@ func (r *ComponentBuildReconciler) cleanupPaCRepositoryIncomings(ctx context.Con
...
 	_ = updateIncoming(repository, incomingSecretName, pacIncomingSecretKey, dummyRevision)
 
+	updateRequired := false
 	// Update repository incomings to only include used revisions
 	if repository.Spec.Incomings != nil && len(*repository.Spec.Incomings) > 0 {
 		// updateIncoming was called above to ensure there's only one entry, so we can safely access index [0]
 		// Set targets to used revisions (can be empty slice)
 		(*repository.Spec.Incomings)[0].Targets = usedRevisions
+		updateRequired = true
+		log.Info("Will update incomings in the PaC repository", "PaCRepositoryName", repository.Name, "Targets", usedRevisions, l.Action, l.ActionUpdate)
```
mmorhun: What is the use for this log?

internal/controller/component_build_controller_pipeline.go
```diff
@@ -806,7 +808,7 @@ func generateCelExpressionForPipeline(component *compapiv1alpha1.Component, gitC
 		pathChangedSuffix = fmt.Sprintf(` && ( "%s***".pathChanged() || ".tekton/%s".pathChanged() %s)`, contextDir, pipelineFileName, dockerfilePathChangedSuffix)
 	}
 
-	return fmt.Sprintf("%s && %s%s", eventCondition, targetBranchCondition, pathChangedSuffix), nil
+	return fmt.Sprintf("%s && %s && %s%s", eventCondition, targetBranchCondition, buildEnabledCondition, pathChangedSuffix), nil
```
mmorhun, line 811: It always adds `builds_enabled == true`, will that mean that the build will happen only when the param set to true? 
Should we use something like `builds_enabled != false` instead?

internal/controller/component_build_controller_pac_repository.go
rcerven: removed

internal/controller/component_build_controller_pipeline.go
rcerven: we are setting that param to either "true" or "false" so there can't be any other value

but I can change that on-cell to  :  builds_enabled != "false"

I don't have preference

mmorhun: Please do, we need to allow builds even if the param is not set (because of old model for example).

mmorhun (approved):

internal/controller/component_build_controller_pipeline.go
mmorhun: I didn't realized that it will fail anyway, so keeping true is fine.

PR #591, opened 2026-04-24
Always use PaC Service to communicate with PaC endpoint from operator

mmorhun:
Always use PaC Service for direct Build Service to PaC communication and configure users' git repositories web hook URL as defined in `webhook-config.json`, if any. 

Also, cache the `webhook-config.json` and don’t re-read. A change in the corresponding config map will cause the deployment change and operator pod rollout anyway.

### DIFF
--- a/internal/controller/component_build_controller_pac.go
+++ b/internal/controller/component_build_controller_pac.go
@@ -51,8 +51,9 @@ const (
 	pipelineRunOnPRSuffix             = "-on-pull-request"
 	pipelineRunOnPushFilename         = "push.yaml"
 	pipelineRunOnPRFilename           = "pull-request.yaml"
-	pipelinesAsCodeNamespace          = "openshift-pipelines"
-	pipelinesAsCodeNamespaceFallback  = "pipelines-as-code"
+	pipelinesAsCodeNamespaceEnvVar    = "PAC_NAMESPACE"
+	pipelinesAsCodeNamespaceOpenshift = "openshift-pipelines"
+	pipelinesAsCodeNamespace          = "pipelines-as-code"
 	pipelinesAsCodeRouteName          = "pipelines-as-code-controller"
 	pipelinesAsCodeRouteEnvVar        = "PAC_WEBHOOK_URL"
 	pipelinesAsCodeWebhooksSecretName = "pipelines-as-code-webhooks-secret"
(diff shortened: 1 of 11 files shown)


rcerven (approved):

internal/controller/component_build_controller_pac.go
```diff
@@ -542,49 +537,54 @@ func (r *ComponentBuildReconciler) UndoPaCProvisionForComponentOldModel(ctx cont
...
+
 	return webhookTargetUrl, nil
 }
 
 // getPaCRoutePublicUrl returns Pipelines as Code public route that recieves events to trigger new pipeline runs.
+// It checks "openshift-pipelines", "pipelines-as-code" namespaces
+// and namespace defined in PAC_NAMESPACE environment variable, if any (takes precedence).
+// The operator should always use this Route when communicating with PaC endpoint.
 func (r *ComponentBuildReconciler) getPaCRoutePublicUrl(ctx context.Context) (string, error) {
 	pacWebhookRoute := &routev1.Route{}
```
gbenhaim, line 566: I think it would have been better to use the kubernetes service of PAC instead of the route, so we will use the same approach both on k8s and Openshift.

If you still prefer to use the route when running on Openshift, please keep the option to override pac's url using the `PAC_WEBHOOK_URL` also for the incoming webhook (and don't check any route in this case). This is essential for keeping the support for k8s.

mmorhun: Yes, you are right, will rework.

internal/controller/component_build_controller_pac.go
```diff
@@ -542,49 +537,84 @@ func (r *ComponentBuildReconciler) UndoPaCProvisionForComponentOldModel(ctx cont
...
 			}
-			// Pipelines as Code public route was not found in expected namespaces
-			// Consider this error permanent
-			return "", boerrors.NewBuildOpError(boerrors.EPaCRouteDoesNotExist,
-				fmt.Errorf("PaC route not found in %s nor %s namespace", pipelinesAsCodeNamespace, pipelinesAsCodeNamespaceFallback))
+			// Not found, continue
+		} else {
+			// Service found, generate URL: http://<service-name>.<namespace-name>.svc.cluster.local
+			internalServiceUrl := fmt.Sprintf("http://%s.%s.svc.cluster.local", pacEndpointServiceKey.Name, pacEndpointServiceKey.Namespace)
+			return internalServiceUrl, nil
```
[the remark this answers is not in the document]
mmorhun: Good catch with port, but I disagree on `svc.cluster.local` part.

rcerven (approved):

gbenhaim (commented):
### Empty-string catch-all default is broken (regression)

The [documented](https://konflux-ci.dev/konflux-ci/docs/guides/webhook-url-configuration/#mixed-per-provider-with-a-default-fallback) catch-all pattern using `""` as a key no longer works:

```yaml
webhookURLs:
  "": "https://smee.example.com/default-channel"
```

In `GetPaCWebhookUrlForGitRepo`, `longestPrefixLen` starts at `0`, so the condition `len(prefix) > longestPrefixLen` evaluates to `len("") > 0` → `0 > 0` → `false`. The empty-string entry is silently skipped.

The old code had an explicit fallback block for this. The fix is to initialize `longestPrefixLen` to `-1`:

```go
func (w *PaCWebhookMapping) GetPaCWebhookUrlForGitRepo(repositoryUrl string) string {
    longestPrefixLen := -1
    matchedTarget := ""
    for prefix, target := range w.mapping {
        if strings.HasPrefix(repositoryUrl, prefix) && len(prefix) > longestPrefixLen {
            longestPrefixLen = len(prefix)
            matchedTarget = target
        }
    }
    return matchedTarget
}
```

This way `0 > -1` is `true`, and a more specific prefix still wins. A unit test for the empty-string catch-all case would be good to add.

---

_This review was performed by Gal's Cursor._

gbenhaim (approved):

PR #669, opened 2026-07-15
feat: component nudging is skipped when in a namespace exists NudgeConfig CR

rcerven:
named "nudge-config", controller will just set "build.appstudio.openshift.io/component-nudge-processed" annotation on pipelineRun and consider it processed.

This is used for migration of nudging to integration service.

[STONEINTG-1672](https://redhat.atlassian.net/browse/STONEINTG-1672)

## Related Issues

[STONEINTG-1672]: https://redhat.atlassian.net/browse/STONEINTG-1672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[KONFLUX-14905]: https://redhat.atlassian.net/browse/KONFLUX-14905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #674, opened 2026-07-20
fix: use main for manifests

tisutisu:
## Description

- When using a specific commit from `konflux-ci/konflux-ci` we were using the same revision of other components in CI, causing CI to fail when we depend on other component's latest changes. So we should use `main` revision so that we get the latest always during konflux deployment.
- And for the konflux deploy task, use latest revision.

## Related Issues

Tests failing on [this PR](https://github.com/konflux-ci/build-service/pull/669)

### DIFF
--- a/e2e-tests/pipelines/konflux-e2e-tests.yaml
+++ b/e2e-tests/pipelines/konflux-e2e-tests.yaml
@@ -98,7 +98,7 @@ spec:
           - name: url
             value: https://github.com/konflux-ci/konflux-ci
           - name: revision
-            value: d91a9ece92d5e0a5ec6408a0a3dc2ec0fa4e0215
+            value: 8a3219a261dca651737a0890c3868c7363af2877
           - name: pathInRepo
             value: .tekton/tasks/deploy-konflux/task.yaml
       params:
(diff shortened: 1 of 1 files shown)


rcerven (approved):

e2e-tests/pipelines/konflux-e2e-tests.yaml
```diff
@@ -98,7 +98,7 @@ spec:
           - name: url
             value: https://github.com/konflux-ci/konflux-ci
           - name: revision
-            value: d91a9ece92d5e0a5ec6408a0a3dc2ec0fa4e0215
+            value: 8a3219a261dca651737a0890c3868c7363af2877
```
rcerven, line 101: shouldn't this be 'main' as well?

rcerven (changes requested):

e2e-tests/pipelines/konflux-e2e-tests.yaml
tisutisu: we want to use a stable version of the task, since any task change could break our CI, so avoided using `main` for the same

rcerven (approved):

